Issue #25, opened 2019-08-07
Bug in S-chess FEN generation


gbtami:
In my up to date Fairy-Stockfish clone `python3 test.py` fails one get_fen() test.
The position is your example in https://github.com/ianfab/Fairy-Stockfish/commit/22fc9678abe3c2fe6d0d23262587dc7a39449abc commit message.
```
tamas@tami:~/Fairy-Stockfish$ python3 test.py
test_get_fen (__main__.TestPyffish) ... FAIL
test_get_san (__main__.TestPyffish) ... ok
test_gives_check (__main__.TestPyffish) ... ok
test_has_insufficient_material (__main__.TestPyffish) ... ok
test_info (__main__.TestPyffish) ... ok
test_is_immediate_game_end (__main__.TestPyffish) ... ok
test_is_optional_game_end (__main__.TestPyffish) ... ok
test_legal_moves (__main__.TestPyffish) ... ok
test_set_option (__main__.TestPyffish) ... info string variant capablanca files 10 ranks 8 pocket 0 template fairy startpos rnabqkbcnr/pppppppppp/10/10/10/10/PPPPPPPPPP/RNABQKBCNR w KQkq - 0 1
ok
test_short_castling (__main__.TestPyffish) ... ok
test_start_fen (__main__.TestPyffish) ... ok

======================================================================
FAIL: test_get_fen (__main__.TestPyffish)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test.py", line 73, in test_get_fen
    self.assertEqual(result, fen1)
AssertionError: 'reb2rk1/ppppqppp/2nb1n2/4p3/4P3/N1P2N2/PB1PQPPP/RE2KBHR[h] w KQacgh - 3 7' != 'reb2rk1/ppppqppp/2nb1n2/4p3/4P3/N1P2N2/PB1PQPPP/RE2KBHR[h] w KQac - 3 7'
- reb2rk1/ppppqppp/2nb1n2/4p3/4P3/N1P2N2/PB1PQPPP/RE2KBHR[h] w KQacgh - 3 7
?                                                                  --
+ reb2rk1/ppppqppp/2nb1n2/4p3/4P3/N1P2N2/PB1PQPPP/RE2KBHR[h] w KQac - 3 7


----------------------------------------------------------------------
Ran 11 tests in 0.005s

FAILED (failures=1)
```
Btw. the bug was reproduced in https://pychess-variants.herokuapp.com/xyF6jgaY

[...]

PR #27, opened 2019-08-07
Fix updating gates on castling move

gbtami:
https://github.com/ianfab/Fairy-Stockfish/blob/master/src/position.cpp#L1300 is wrong because when we are doing a castling move gates can't be set on `to` square if it's a standard game.
If it's a Chess960 castling (king takes rook) the gates on `to` square may or may not be set depending on if we have something in hand to gate or not.

On the other side castling rook `from` square have to be cleared in gates.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -1063,8 +1063,6 @@ void Position::do_move(Move m, StateInfo& newSt, bool givesCheck) {
   Square to = to_sq(m);
   Piece pc = moved_piece(m);
   Piece captured = type_of(m) == ENPASSANT ? make_piece(them, PAWN) : piece_on(to);
-  Square rfrom;
-
   if (to == from)
   {
       assert(type_of(m) == PROMOTION && sittuyin_promotion());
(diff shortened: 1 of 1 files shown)


gbtami:
This fixes https://github.com/ianfab/Fairy-Stockfish/issues/25

src/position.cpp
```diff
@@ -1296,8 +1298,8 @@ void Position::do_move(Move m, StateInfo& newSt, bool givesCheck) {
   {
       if (is_ok(from) && (gates(us) & from))
           st->gatesBB[us] ^= from;
-      if (type_of(m) == CASTLING)
-          st->gatesBB[us] ^= to;
+      if (type_of(m) == CASTLING && (gates(us) & rfrom))
```
ianfab: Thanks, looks good to me. Just as a minor note, in order to reduce the frequency of merge conflicts from upstream I usually try to avoid touching upstream code if feasible. Here one could use `to_sq(m)` instead of `rfrom` in order to retain the scope of `rfrom`, since SF always uses the rook square as `to_sq` in move encoding (irrespective of whether it is 960 or not), although `rfrom` might be more explicit.

gbtami: Thx! Done.

ianfab:
Thx, merged in https://github.com/ianfab/Fairy-Stockfish/commit/0f48768c5ecbc705f4a92096598ad748eaccc7f6.

PR #44, opened 2019-11-02
Support XBoard protocol (close #37)

### LINKED ISSUE #37: XBoard protocol support
Instead of relying on the [uci2wb adapter](https://github.com/ianfab/uci2wb), it would be preferable to have built-in XBoard/CECP protocol support in order to have full control over the engine-GUI communication.

Experimental branch: https://github.com/ianfab/Fairy-Stockfish/tree/xboard

References:
http://home.hccnet.nl/h.g.muller/interfacing.txt
https://www.gnu.org/software/xboard/engine-intf.html#8

ianfab:
Implements CECP/XBoard as third supported protocol besides UCI and USI.

Since I am not an expert on the CECP protocol, independent testing or review is very welcome. So far I tested analysis and game play with XBoard and cutechess for a few variants, which seems to be working fine.

Implementation is based on specification in https://www.gnu.org/software/xboard/engine-intf.html#8.

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -228,9 +228,17 @@ void MainThread::search() {
   {
       rootMoves.emplace_back(MOVE_NONE);
       Value variantResult;
+      Value result =  rootPos.is_game_end(variantResult) ? variantResult
+                    : rootPos.checkers()                 ? rootPos.checkmate_value()
+                                                         : rootPos.stalemate_value();
+      if (Options["Protocol"] == "xboard")
+          sync_cout << (  result == VALUE_DRAW ? "1/2-1/2 {Draw}"
+                        : (rootPos.side_to_move() == BLACK ? -result : result) == VALUE_MATE ? "1-0 {White wins}"
+                        : "0-1 {Black wins}")
+                    << sync_endl;
+      else
       sync_cout << "info depth 0 score "
-                << UCI::value(  rootPos.is_game_end(variantResult) ? variantResult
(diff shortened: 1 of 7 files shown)


alwey:
I tested the PR version and it worked very well (for standard chess and mini shogi). Minor issue: I noticed that the mate-scores in xboard-mode are not quite compatible with the specs, e.g. giving a score of 500.03 instead of 499.97 (old spec) or 1000.03 (new spec).

ianfab:
Thanks, I will check the mate score issue.

ianfab:
Lol, first I though I had a sign error, but then I noticed I overlooked a division by two. Anyhow, it should be fixed now.

Do you by chance know whether there is any (well-known) GUI that does not send the time/otim commands to sync the clocks? I am asking because SF does not keep track of the time at all, it just relies on that the GUI always sends the time command.

alwey:
AFAIK fairly recent UIs that support Xboard protocol V2 will send `time` / `otim` if `feature time=1`, which is the default.  Perhaps it would be good to issue `feature .... time=1` explicitely.

The initial mechanism for time control is based on the `level` command and Xboard-engines were expected to have their own time measurement, especially protocol V1 engines. So this is an example of the fact that UCI engines are simpler and Xboard engines are more complex.

ianfab:
Since time=1 is the default according to the spec, I did not include it, but I can of course add it to be safe. I just tried to keep the xboard state machine as small as possible, so if all protocol compliant GUIs send the time command when enabled, then it should be fine.

PR #84, opened 2020-03-05
WASM build

easychessanimations:
I submit this pull request not with the intent of merging, but to show what I have in creating a WASM build for Fairy-Stockfish.

I saved the original Makefile in `Makefile_orig` and created a new `Makefile_wasm`, which should be copied to Makefile.

The only source file minimally modified is `parser.cpp`, because `mingw` complained about `std::stringstream` being not copyable.

The build command is:
```
make ARCH=wasm build -j
```

To test the build, run `http-server` standing in the `src` directory, open the browser console, then visit `localhost:8080`. The program loads but freezes after printing the welcome message.


easychessanimations:
Added a commit which replaces the UCI loop with a command interpreter function that can react to Javascript worker messages, avoids threads startup and adds more logging. With these changes the program runs and executes the `uci` command properly. In the lack of threads it won't analyze.

ianfab:
Thanks, sounds good. Looking at the diff between stockfish.wasm and stockfish I see the following changes you did not port yet:
* Removal of tablebases
  * I actually do not understand why this should be required. If we want to prevent issues with file system access in WASM (since it normally does not have access) we can simply return early in `Tablebases::init` like it is done for the large-board version and the engine will never try to load TB files, which is way simpler and has basically the same effect as removing the code.
* Reallocation of threads and hash
  * To my understanding this is just an optimization in stockfish.wasm and is not required to make it work. If this is the case I would leave it out for now unless there is a significant impact on startup time or the like.
* Thread startup
  * It is mentioned that there would be a deadlock without the changes to the thread startup in stockfish.wasm, have you observed this?
* Update constraints on UCI options
  * WASM seems to put constraints on threads and memory, so reflecting this in the UCI option limits perhaps makes sense.
* Enforcing usage of custom `NativeThread` implementation for emscripten
  * I do not know whether this is required, but it is anyway only a one-line change. We should however definitely not reduce the stack size, since that would crash Fairy-Stockfish.

smartchessguiapp:
If I try thread startup, then it freezes. I discovered this by adding logging to every initalization step in `main.cpp` and it froze at threads init. Without threads startup, search also could not be cleared, so I had to remove both to reach `post run` state, meaning that main returned and the worker is ready to execute commands.

easychessanimations:
If I try threads startup then it freezes. I discovered this by adding logging statements after every initialization step, and it froze at threads startup. Without threads stratup clear search also freezes, so I had to remove both to reach `post run` status, meaning that `main` has returned and the worker is up and ready to execute commands.

PR #88, opened 2020-03-27
Increase king danger for makpong

Fulmene:


ianfab:
Thanks!

By the way, you can also use the domain URL instead of IP: http://www.variantfishtest.org:6543/tests

PR #89, opened 2020-03-29
Increase leapers' value for makpong

Fulmene:


ianfab:
In principle looks good to me. However, I would prefer to make the condition that the leap is more than one square more explicit, e.g., by introducing an `int distance(Direction d)` function (e.g., in types.h) and then checking for distance(d) >= 2 instead of checking against all the specific directions. Converting `Direction` to a distance is not trivial, but something like in http://cpp.sh/3ec2e should work.

Fulmene:
I see. That would help the code looks more readable. I'll verify distance conversion again also.

Do you want me to include it in this PR?

ianfab:
Would be nice if you could include it in this PR if that is fine with you, otherwise I could also do the refactoring afterwards.

ianfab:
Looks good. Can you please rebase and squash into a single commit (containing the test results in the message)?

ianfab:
I rebased the branch on top of master. I will wait for the CI to finish, then it should be ready for merge.

ianfab:
Congrats and thanks again, it is nice to see some progression in playing strength.

PR #114, opened 2020-04-26
Support specifying the start of counting phase

Fulmene:


src/position.cpp
```diff
@@ -686,7 +686,7 @@ const string Position::fen(bool sfen, bool showPromoted, std::string holdings) c
 
   // Counting ply or 50-move rule counter
   if (st->countingLimit)
-      ss << st->countingPly;
+      ss << std::min(st->countingPly, countStarted > (unsigned int)gamePly ? 0 : (int)(1 + gamePly - countStarted));
```
ianfab, line 689: A few things here, also applies to the equivalent line in `is_optional_game_end`:
1. The case of countStarted == 0 with gamePly < countingPly does not seem to be considered yet. In case of countStarted == 0 the result should always be countingPly.
2. I do not see where the +1 (on top of gamePly) is coming from. What is the reason for it?
3. I slightly prefer using std::max over the ternary operator in such a case, since the capping of the value is more descriptive then.

test.py
```diff
@@ -216,6 +216,43 @@ def test_get_fen(self):
         result = sf.get_fen("makruk", fen, [], False, False, True)
         self.assertEqual(result, fen)
 
+        # makruk piece honor counting
+        fen = "8/3k4/8/2K1S1P1/8/8/8/8 w - - 0 50"
+        moves = ["g5g6M"]
+        result = sf.get_fen("makruk", fen, moves, False, False, True)
+        self.assertEqual(result, "8/3k4/6M~1/2K1S3/8/8/8/8 b - 88 8 50");
```
ianfab, line 223: Here and below the trailing semicolons could be removed.

test.py
```diff
@@ -216,6 +216,43 @@ def test_get_fen(self):
...
+        fen = "3k4/2m5/5M~2/4M3/3KS3/8/8/8 w - 128 0 33"
+        moves = ["d4d5"]
+        result = sf.get_fen("makruk", fen, moves, False, False, True)
+        self.assertEqual(result, "3k4/2m5/5M~2/3KM3/4S3/8/8/8 b - 128 1 33");
+
+        fen = "3k4/2m5/5M~2/4M3/3KS3/8/8/8 w - 128 0 33"
+        moves = ["d4d5"]
+        result = sf.get_fen("makruk", fen, moves, False, False, True, 0)
+        self.assertEqual(result, "3k4/2m5/5M~2/3KM3/4S3/8/8/8 b - 128 1 33");
+
```
ianfab, line 245: It would be good to also cover the above mentioned scenario of gamePly < countingPly (with countStarted == 0) in the tests.

test.py
```diff
@@ -216,6 +216,43 @@ def test_get_fen(self):
         result = sf.get_fen("makruk", fen, [], False, False, True)
         self.assertEqual(result, fen)
 
+        # makruk piece honor counting
+        fen = "8/3k4/8/2K1S1P1/8/8/8/8 w - - 0 50"
+        moves = ["g5g6M"]
```
ianfab, line 221: Officially in the protocol the promotion piece type is required to be lowercase, uppercase is just allowed for compatibility with some non-compliant GUIs.

src/position.cpp
Fulmene: 1. Okay. Will do.
2. The countStarted number is the game ply where the count was at one. Therefore, if countStarted and gamePly are equal, the result should be 1. Because 0 is reserved as the "no counting" value, 1 needs to represent the earliest a player can theoretically start counting, which is game ply 1, the first move of the game by White.
3. I think using std::max in this case will be problematic. Since the number typically used for disable counting is -1, 1 + gamePly - countStarted will be equal to 2 + gamePly in that case, and std::max won't give the desired 0 result.

Fulmene: For the check in `is_optional_game_end`, which would you prefer?
```
std::min(st->countingPly, countStarted > (unsigned int)gamePly ? 0 : countStarted == 0 ? st->countingPly : (int)(1 + gamePly - countStarted)) > st->countingLimit
```
VS
```
((countStarted == 0 && st->countingPly > st->countingLimit)
          || std::min(st->countingPly, countStarted > (unsigned int)gamePly ? 0 : (int)(1 + gamePly - countStarted)) > st->countingLimit)
```

ianfab: Thanks, I missed the overflow issue. IMO zero would not necessarily need to be a reserved value as activation on ply zero to me means the same as always enabled, but it is also fine the way it is now. I will probably merge it later today (CET).

Fulmene:
Should I squash these changes into a single commit?

Issue #157, opened 2020-06-25
JavaScript Library Binding


QueensGambit:
Thank you @ianfab for your wonderful work on _Fairy-Stockfish_.

What is your opinion on adding a JavaScript library binding, **ffish.js**, for _Fairy-Stockfish_?

I have seen that you have already created a [web assembly fork](https://github.com/ianfab/stockfish.wasm) for _Fairy-Stockfish_.
However, this version is only meant to be used as an engine and does not allow move generation, checking for valid FENs etc., as an engine is supposed to.

I know that there exists [chess.js](https://github.com/jhlywa/chess.js) and even a dedicated fork for crazyhouse ([crazyhouse.js](https://github.com/jak3122/crazyhouse.js)).
Unfortunately, [crazyhouse.js](https://github.com/jak3122/crazyhouse.js) is limited to crazyhouse, the development has stopped and it isn't optimized.

I am asking because I started to create an UCI chess variant analysis web GUI which is meant to be later exported to the deskop using [electron.js](https://www.electronjs.org/). This way you can integrate the infamous [lichess.org chessboard](https://github.com/gbtami/chessgroundx) and other powerful web assets while still running the engine executables natively with multi-processing and possibly GPU access via [node.js](https://nodejs.org/en/).

![image](https://user-images.githubusercontent.com/15967624/85744678-5e1e8780-b705-11ea-81c2-263e4a75f1a4.png)
_Layout of the GUI (WIP)_

I'm planning to open source this GUI after I created the first release candidate. Some open-source chess engines like [Muti-Variant-Stockfish](https://github.com/ddugovic/Stockfish) and [Fairy-Stockfish](https://github.com/ianfab/Fairy-Stockfish) are expected to be integrated as standard engines in this GUI.
Currently, it is using [crazyhouse.js](https://github.com/jak3122/crazyhouse.js) as its back-end. Alternatively, I could add a python back-end using [pyffish](https://pypi.org/project/pyffish/) or [python-chess](https://pypi.org/project/python-chess/) but I would prefer avoiding an additional overhead.

[...]

Issue #87, opened 2020-03-20
FEN validation for pyffish and ffishjs


ianfab:

ianfab:
Open:
* [ ] Expose FEN validation in pyffish
* [ ] Refactor FEN validation to better align with variant configuration properties

gbtami:
Sry for the late join, but I have some more suggestion we implemented in pychess.

Decline FEN if the king is in check, but the moving side is the opponent (in variants where king can't be captured)

Similar to how FEN with missing parts handled, it would be useful to accept drop variant FEN with no empty bracket pairs at all.

It would also be nice to accept lichess style crazyhouse FEN as well.

ianfab:
Lichess style crazyhouse FENs are already supported. For the other ones I opened a new issue. Feel free to comment there if you come up with more scenarios.

PR #158, opened 2020-06-27
JavaScript Chess Variant Library: ffish.js

QueensGambit:
Hello @ianfab ,

I created the initial version of **ffish.js** today, as discussed in https://github.com/ianfab/Fairy-Stockfish/issues/157.

The main functionality such as creating a new board from fen, loading a new fen, generating all legal moves (UCI or SAN) is covered. The syntax of **ffish.js** is similar to [python-chess](https://github.com/niklasf/python-chess).

Moreover, I moved some utility functions which are both used in **pyffish** and **ffish.js** to a new file called `apiutil.h`.

## Build instuctions

```bash
cd Fairy-Stockfish/src
```
```bash
emcc -O3 --bind ffishjs.cpp \
benchmark.cpp \
bitbase.cpp \
bitboard.cpp \
endgame.cpp \
evaluate.cpp \
material.cpp \
misc.cpp \
movegen.cpp \
movepick.cpp \
parser.cpp \
partner.cpp \
pawns.cpp \
piece.cpp \
position.cpp \
psqt.cpp \
search.cpp \
thread.cpp \
timeman.cpp \
tt.cpp \
uci.cpp \
syzygy/tbprobe.cpp \
ucioption.cpp \
variant.cpp \
xboard.cpp \
-o ../tests/js/ffish.js
```

## Exemplary usage in JavaScript

Load the API in JavaScript:

```javascript
const ffish = require('./ffish.js');
```

Create a new variant board from its default starting position:

```javascript
let board = new ffish.Board("chess");
```

Set a custom fen position:
```javascript
board.setFen("rnb1kbnr/ppp1pppp/8/3q4/8/8/PPPP1PPP/RNBQKBNR w KQkq - 0 3");
```

Initialize a board with a custom FEN:
```javascript
board = new ffish.Board("crazyhouse", "rnb1kb1r/ppp2ppp/4pn2/8/3P4/2N2Q2/PPP2PPP/R1B1KB1R/QPnp b KQkq - 0 6");
```

Pushing a new move to the board:
```javascript
board.push("g2g4");
```

Generate all legal moves in UCI and SAN notation:
```javascript
let legalMoves = board.legalMovesStr().split(" ");
let legalMovesSan = board.legalMovesSanStr().split(" ");

for (var i = 0; i < legalMovesSan.length; i++) {
    console.log(`${i}: ${legalMoves[i]}, ${legalMovesSan[i]}`);
}
```

## Performance

I conducted some initial performance test in comparision with [**chess.js**](https://github.com/jhlywa/chess.js) and [**crazyhouse.js**](https://github.com/jak3122/crazyhouse.js).

### Move generation of two example positions for 1k iterations
```javascript
Standard Chess
==================
Call to board.legalMoves() took 27.46  milliseconds.
Call to board.legalMovesStr()+legalMoves.split(" ") took 24.84  milliseconds.
board.legalMovesSanStr().split(" ").length: 29
Call to board.legalMovesSanStr()+legalMoves.split(" ") took 23.40  milliseconds.
chess.moves().length: 29
Call to chess.moves() took 880.41  milliseconds.
Crazyhouse
===========
board.legalMovesStr().split(" ").length: 97
Call to board.legalMoves() took 55.29  milliseconds.
crazyhouse.moves().length: 97
Call to crazyhouse.moves() took 10257.35  milliseconds.
```

**ffish.js** is **~35x** faster than **chess.js** and **~185x** faster than **crazyhouse.js** for SAN move generation.

You can find the code of the testing server here:
* https://github.com/QueensGambit/Fairy-Stockfish/tree/ffish_js/tests/js

Currently, there are two versions for generating the moves.

`board.legalMovesSan()` and `board.legalMovesSanStr()`. The former returns the moves as a vector of strings and the later as a concatenated string with whitespaces.
Surprisingly, the concatenated string plus a splitting operation is faster than the vector form as seen above.
If you agree, I suggest to remove the vector form then.

Another thing which I found odd, is that `board.legalMovesSanStr()` is slightly faster than generating the moves in UCI form by using `board.legalMovesStr()`.

## TODOs

There are still few important TODOs:
* Adding a `pushSan()` function which requires converting a SAN move to _Fairy-Stockfish's_ move representation
* Validating FEN strings
* Loading PGN files

I can imagine that a native _Stockfish_ or _Multi-Variant-Stockfish_ version **fish.js**/**mvfish.js** for standard chess is 15x/10x faster than **ffish.js**.

It might be interesting to run a performance comparision with the [lichess-Scala-API](https://github.com/ornicar/scalachess) then.
I saw that you also added a bughouse backend to _Fairy-Stockfish_. Lichess still doesn't support boughouse, so it could benefit from accessing your API via **ffish.js**.


ianfab:
Wow, that is awesome. The performance also sounds very promising, although I agree that most likely a pure fish.js version could perhaps be ~10x faster as also perft is ~10x faster, but considering the already quite good performance compared to other libraries it probably will not be the bottleneck anyway. Below I have some more feedback:

- The addition of the `ffishjs.cpp` breaks the existing appveyor builds
  - The MSVC build must exclude the new file, see https://github.com/ianfab/Fairy-Stockfish/blob/f712f6f2d455ebe570e406edc369d332d5d655ab/appveyor.yml#L35
  - The python extension also needs to exclude the new source file to work, see https://github.com/ianfab/Fairy-Stockfish/blob/f712f6f2d455ebe570e406edc369d332d5d655ab/setup.py#L27-L30
- For source files that are purely Fairy-Stockfish code and do not originate from official SF (as for apiutil.h) I usually use a different copyright header, see e.g., variant.h or the like.
- Including the building with emscripten and ideally also a few tests in the CI would be nice to avoid future regressions.
- SF source code usually uses the convention to have opening braces for if/for (not for functions, methods, or classes) on new lines, and to skip them if not required (for single statements).
- A simple stupid implementation of pushSan could be to just generate the list of legal SAN moves and compare the given string to them. This would of course neither be a very robust parsing (e.g., failing on optional disambiguation) nor the most performant, but might be sufficient for now.
- FEN validation is planned since it is also required for pyffish and the CECP protocol, see https://github.com/ianfab/Fairy-Stockfish/issues/87 (but I have not started on it yet).

The first point is the only one I see as a must to include in this PR, everything else can be optional and also be done at a later point.

QueensGambit:
I addressed your points.

* I excluded the `ffishjs.cpp` from the `pyffish` build.

* The license of `apiutil.h`was updated and the following license header was added to `ffishjs.cpp`:
```
/*
  ffish.js, a JavaScript chess variant library derived from Fairy-Stockfish
  Copyright (C) 2020 Fabian Fichter, Johannes Czech
  ffish.js is free software: you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
  the Free Software Foundation, either version 3 of the License, or
  (at your option) any later version.
  ffish.js is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  GNU General Public License for more details.
  You should have received a copy of the GNU General Public License
  along with this program.  If not, see <http://www.gnu.org/licenses/>.
*/
```
If you want we can choose a different license here.

I also added some new wrapper function:
* Board()
* Board(std::string uciVariant, std::string fen, bool is960)
* pushSan()
* pop()
* reset()
* is960()
* turn()
* gamePly()
* isGameOver()

Renamed / replaced:
* legalMovesStr to legalMoves
* legalMovesSanStr to legalMovesSan


I added a unit test for every method using [mocha](https://mochajs.org/):

```js
npm test

  Constructor: no parameter 
    ✓ it creates a chess board from the default position (98ms)

  Constructor: variant parameter 
    ✓ it creates a board object from a given UCI-variant

  Constructor: variant parameter + fen 
    ✓ it creates a board object for a given UCI-variant with a given FEN

  Constructor: variant parameter + fen + is960
    ✓ it creates a board object for a given UCI-variant with a given FEN and is960 identifier

  board.legalMoves()
    ✓ it returns all legal moves in uci notation as a concatenated string

  board.legalMovesSan()
    ✓ it returns all legal moves in SAN notation as a concatenated string

  board.numberLegalMoves()
    ✓ it returns all legal moves in uci notation as a concatenated string

  board.push()
    ✓ it pushes a move in uci notation to the board

  board.pushSan()
    ✓ it pushes a move in san notation to the board

  board.pop()
    ✓ it undos the last move

  board.reset()
    ✓ it resets the board to its starting position

  board.is960()
    ✓ it checks if the board originates from a 960 position

  board.fen()
    ✓ it returns the current position in fen format

  board.setFen()
    ✓ it sets a custom position via fen

  board.sanMove()
    ✓ it converts an uci move into san

  board.turn()
    ✓ it returns the side to move

  board.halfmoveClock()
    ✓ it returns the halfmoveClock / 50-move-rule-counter

  board.gamePly()
    ✓ it returns the current game ply

  board.isGameOver()
r1bqkb1r/pppp1Qpp/2n2n2/4p3/2B1P3/8/PPPP1PPP/RNB1K1NR b KQkq - 0 4
    ✓ it checks if the game is over based on the number of legal moves


  19 passing (260ms)
```
The corresponding `appveyor.yml` file needs to be created to integrate into CLI.

Somehow the CI is still failing.

If version 0.1 is done we could publish it on [npm](https://www.npmjs.com/).

appveyor.yml
```diff
@@ -32,7 +32,7 @@ init:
 before_build:
   - ps: |
       # Get sources
-      $src = get-childitem -Path *.cpp -Recurse -Exclude pyffish.cpp | select -ExpandProperty FullName
+      $src = get-childitem -Path *.cpp -Recurse -Exclude pyffish.cpp ffish.cp | select -ExpandProperty FullName
```
ianfab, line 35: There are typos in `ffishjs.cpp`. And to my knowledge it requires the patterns as comma-separated, that is where the cryptic CI error message might come from.

setup.py
```diff
@@ -24,9 +24,12 @@
 with io.open("Readme.md", "r", encoding="utf8") as fh:
     long_description = fh.read().strip()
 
+sources = glob("src/*.cpp") + glob("src/syzygy/*.cpp")
+sources.remove('src/ffishjs.cpp')
```
ianfab, line 28: I think this `remove` does not work on Windows (which is why appveyor is failing) since although the glob has to use Unix-style paths, the returned lists of paths are OS specific. So using `os.path.join` might work, or if you want to be on the save side you can use a glob again to definitely get the same path.

QueensGambit:
I added another commit https://github.com/QueensGambit/Fairy-Stockfish/commit/589fa3d13dba0fea5c53eff297ced7710399d40e to the **ffish_js** branch which tries to fix the pyffish CLI but GitHub doesn't seem to update the PR properly. Maybe it has something to do with the current incidents of GitHub:
* https://www.githubstatus.com/incidents/d60fmldlvw48?utm_ts=1593423886

ianfab:
Sorry, I wanted to comment the solution here hours ago already, but due to github incidents I was not able to.

I missed that the glob does not expand since it is a fixed path, so it remains with `/`. I now checked on Windows that using `os.path.join` works fine when constructing a path and removing it from a list of paths of a glob.

QueensGambit:
Should I create a new PR or can you push this change to the PR?
As mentioned, the PR wasn't updated properly.

ianfab:
I assume you are reffering to https://github.com/QueensGambit/Fairy-Stockfish/commit/589fa3d13dba0fea5c53eff297ced7710399d40e. However, I think this will not fix it, since normcase does not do the `/`->`\\` separator change of the path for windows. If the PR does not update, then maybe squashing and force pushing might trigger it.

ianfab:
The force push indeed seems to have triggered it. From my perspective it looks fine, I will just wait for the CI to succeed and will then probably merge it in the evening. However, since a squash merge from my side would make me the author of the commit it would be nice if you could squash all your changes into one commit before I merge to retain the author information. Thanks a lot again for your contribution.

PR #161, opened 2020-07-04
xboard: implement 'myversion' option

ydirson:


ydirson:
Hm, please ignore this one, will resubmit a more useful one

PR #162, opened 2020-07-04
xboard: add engine version to 'myname' feature

ydirson:


ianfab:
Thanks, including the version sounds like a good idea. In order to simplify merging from official Stockfish, I usually try to make as few changes to upstream code as possible, therefore I think it would be good to e.g. just add an additional argument (like `to_xboard`) to the function and then skip the author information when it is enabled, similar to the `to_uci` parameter.

ydirson:
I was going to suggest using an enum instead of 2 bools that cannot be true at the same time, but as `to_uci` exists upstream, `to_xboard` looks like the simplest.

ydirson:
Oh, while we're talking, and since I don't see any other obvious channel to contact you...

I've packaged Fairy-Stockfish for inclusion in Debian, it should appear at https://tracker.debian.org/pkg/fairy-stockfish, hopefully within a few days.

ianfab:
Thanks. The change looks good, I will just wait for CI to finish and then merge it.

PR #174, opened 2020-08-23
Updated ffish.js for initial npm version

QueensGambit:
Hello @ianfab ,

due to a request by @gbtami who wants to use the [ffish.js library](https://github.com/ianfab/Fairy-Stockfish/tree/master/tests/js) for the [pychess-variants server](https://www.pychess.org/), I created an official [npm-package](https://www.npmjs.com/package/ffish) for it. The library has an unpacked size of 692 kB including the wasm file and 0 dependencies.

Moreover, I updated [package.json](https://github.com/QueensGambit/Fairy-Stockfish/blob/ffish_js/tests/js/package.json) accordingly and added a `ffish.info()` method which returns the version of the _Fairy-Stockfish_ binary.

I discovered that some js frameworks like [vue-js](https://vuejs.org/) require a wasm library to be exported as an [ES2015](https://en.wikipedia.org/wiki/ECMAScript) module ([emscripten/#10114](https://github.com/emscripten-core/emscripten/issues/10114)).
Therefore, I added additional compile instructions to [README.md](https://github.com/QueensGambit/Fairy-Stockfish/tree/ffish_js/tests/js#compile-as-es6es2015-module) to achieve this.
However, I was unsuccessful in loading the ES2015 module in native node.js where ES2015 is still experimantal.

The current version published on npm (0.1.4) is the regular version and uses the regular syntax to load the library.
```javascript
const ffish = require('ffish');
```

Maybe we can make both versions available on npm but this will increase maintenance effort or we find a proper way at exporting the ES2015 version for node.js.

### DIFF
--- a/src/ffishjs.cpp
+++ b/src/ffishjs.cpp
@@ -206,10 +206,16 @@ private:
   }
 };
 
+// returns the version of the Fairy-Stockfish binary
+std::string info() {
+   return engine_info();
+}
+
 bool Board::sfInitialized = false;
 
 // binding code
 EMSCRIPTEN_BINDINGS(ffish_js) {
+  function("info", &info);
   class_<Board>("Board")
     .constructor<>()
     .constructor<std::string>()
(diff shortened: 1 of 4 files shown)


ianfab:
Thanks!

Issue #160, opened 2020-06-28
Evaluation of dead gating pieces in S-Chess


gbtami:
In game https://www.pychess.org/JtHBCNCQ after back move 21... Exd1+ it evaluates 22. Rxd1 more than +2 at high depths because it thinks an elephant in the pocket is something useful. But after Rxd1 white can't gate his elephant anymore! Maybe fixing this can gain some Elo in S-chess :)

ianfab:
Thanks. Yes, this is a known issue and was also partially discussed in #135, but unfortunately I so far failed to find a patch that gains Elo by fixing this, e.g., http://www.variantfishtest.org:6543/tests/view/5ecd66b56e23db36d55f2c49, although it would seem like this should be a no-brainer.

ianfab:
@alwey Congrats, http://www.variantfishtest.org:6543/tests/view/5f5368236e23db221d9e90df passed at STC, feel free to submit for LTC.

alwey:
Thank you. I'll submit soon. Currently I am investigating the midgame weight (200 in the tested version, 700 may be reasonable, too).

ianfab:
@alwey Your test by accident is for standard chess. Tip: To submit a passed STC test for LTC quickly and reliably, you use the "reschedule" button on the finished test, then click on the "LTC" button, and then "submit".

alwey:
Ah thanks, I have not seen that!

alwey:
LTC passed. The patch uses (200,900) as piece score. Perhaps it would be better to scale the actual values of the pieces (this is more general):
```
if (pos.seirawan_gating() && !pos.piece_drops() && pos.count_in_hand(Us, ALL_PIECES) > popcount(pos.gates(Us)))
    score -= make_score(PieceValue[MG][pt] / 10, PieceValue[EG][pt] * 2 / 5)
           / pos.count_in_hand(Us, ALL_PIECES) * (pos.count_in_hand(Us, ALL_PIECES) - popcount(pos.gates(Us)));
```

ianfab:
Nice. Yes, I agree that it would be nice to use a more general formula that considers the piece value. However, securing the Elo gain from a passed patch by merging it should be the priority, after that it can still be simplified, generalized, or improved, then of course testing against the passed patch, having the Elo gain already in the pocket.

alwey:
OK, I just created a PR.

ianfab:
[...]

PR #181, opened 2020-09-06
Evaluate ungatable pieces in Seirawan

alwey:
Add a penalty for pieces in hand if there are not enough gates left.


ianfab:
Thanks and congrats! Closes #160.

For reference:
http://www.variantfishtest.org:6543/tests/view/5f5368236e23db221d9e90df
http://www.variantfishtest.org:6543/tests/view/5f54b6e16e23db221d9e90f5

PR #182, opened 2020-09-08
Added functionality to ffish.js (closes #177)

### LINKED ISSUE #177: Missing ffishjs features
Some more Fairy stuff would be useful to be exposed for pychess client side analysis:
- [ ] setoption(name, value) to set variants.ini path
- [ ] Board.isCheck() to style king in check
- [ ] Board.isBikjang() to style kings in bikjang
- [ ] Board.movesToSan(movelist, notation) to convert analysis PV to SAN movelist
- [ ] Board.san_move() should accept notation parameter also (now NOTATION_SAN is hardwired)

QueensGambit:
Hello @ianfab  and @gbtami ,

I added the requested functionality from #177 resulting in **ffish.js v0.2.0**.

- [x] setoption(name, value) to set variants.ini path
Is now available as a static method. Because you can only overload methods based on the number of arguments from emscripten and not based on variable type, I exported additional methods for integer and boolean options.
However, loading files from JavaScript is tricky and unavailable for browsers.
It seems you require the [filesystem-API](https://emscripten.org/docs/api_reference/Filesystem-API.html#including-file-system-support) for that because I receive: `Unable to open file variants.ini` when setting the `VariantPath`.
Perhaps it is better to add an additional method just for parsing a `variant.ini`-string without reading the file within the library.
  - ffish.setOption()
  - ffish.setOptionInt()
  - ffish.setOptionBool()

- [x] Board.isCheck() to style king in check
  - board.isCheck()

- [x] Board.isBikjang() to style kings in bikjang
  - board.isBikjang()

- [x] Board.movesToSan(movelist, notation) to convert analysis PV to SAN movelist
I chose the name `variationSan()` instead, same as in [python-chess](https://python-chess.readthedocs.io/en/latest/core.html#chess.Board.variation_san) for consistency.
However, it has an additional optional parameter `moveNumbers` which allows to disable move numbers.
  - board.variationSan(uciMoves)
  - board.variationSan(uciMoves, notation)
  - board.variationSan(uciMoves, notation, moveNumbers)
- [x] Board.san_move() should accept notation parameter also (now NOTATION_SAN is hardwired)
I exported the enum Notation which can now be used as an optional parameter.
  - board.sanMove(notation)
  - enum ffish.Notation

- [x] ES6 module export
  - I published [ffish-es6](https://www.npmjs.com/package/ffish-es6) as a separate module to npm. Moreover, I added badges to [README.md](https://github.com/ianfab/Fairy-Stockfish/blob/master/tests/js/README.md) to track how often each package is downloaded and to see which one is more popular.

### Additional method(s)

- [x] board.fullmoveNumber()
  Returns the full move counter starting at 1.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -5,6 +5,7 @@
 [![Build Status](https://travis-ci.org/ianfab/Fairy-Stockfish.svg?branch=master)](https://travis-ci.org/ianfab/Fairy-Stockfish)
 [![Build Status](https://ci.appveyor.com/api/projects/status/github/ianfab/Fairy-Stockfish?branch=master&svg=true)](https://ci.appveyor.com/project/ianfab/Fairy-Stockfish/branch/master)
 [![PyPI version](https://badge.fury.io/py/pyffish.svg)](https://badge.fury.io/py/pyffish)
+[![NPM version](https://img.shields.io/npm/v/ffish.svg?sanitize=true)](https://www.npmjs.com/package/ffish)
 
 Fairy-Stockfish is a chess variant engine derived from [Stockfish](https://github.com/official-stockfish/Stockfish/) designed for the support of fairy chess variants and easy extensibility with more games. It can play various regional, historical, and modern chess variants as well as [games with user-defined rules](https://github.com/ianfab/Fairy-Stockfish/wiki/Variant-configuration). For [compatibility with graphical user interfaces](https://github.com/ianfab/Fairy-Stockfish/wiki/Usage) it supports the UCI, UCCI, USI, and CECP/XBoard protocols.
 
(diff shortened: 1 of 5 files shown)


gbtami:
Regarding reading variants.ini file I have faced the similar problem in https://github.com/ianfab/stockfish.wasm/issues/4
As @ianfab mentioned the reading itself should work if the INI file packaged at compile time. Unfortunately I get errors using --preload-file and --embed-file  also :(

ianfab:
Thanks!

Yes, regarding the loading of variant configurations it would make sense to enable the js and py bindings to also be able to directly pass a string instead of a file path (it could also be a more direct API to define variants, but that might be overkill). Generalizing the current method using an `ifstream` to allow any `istream` (`ifstream`/`stringstream`) should be straightforward.

PR #190, opened 2020-10-04
[WIP] add nightrider piece

yoav-rozin:
Hey,
I have tried to add new pieces (in this case the nightrider). The program compiled successfully, but I have a problem: I can't move with the new piece (Error (unkown command)).
I was trying to add the nightrider based on the references of biskni piece, and the only reference I found and I didn't add is the "value", is it what I need to add?
I'll be glad if you can explain me what did I miss.
Thanks.


ianfab:
Thanks for your suggestion. Unfortunately it is not that easy to implement the nightrider, since new types of riders need to be implemented specifically, see e.g. https://github.com/ianfab/Fairy-Stockfish/blob/master/src/bitboard.cpp. Therefore, specifying a direction in the sliderQuiet/sliderCapture does not have any effect as long as the corresponding rider type is not implemented. Nightriders might also need some additional special handling due to their leaps (compared to the normal sliders), which, besides the potential performance impact, is one of the reasons why I so far refrained from trying to implement it.

Yes, the piece value would also need to be added, but that is just relevant for the evaluation, it does not have any effect on the move generation.

Issue #81, opened 2020-02-28
Atomic / Atomic-960


smartchessguiapp:
Dear ianfab !

I admire your work about which I learned at https://www.pychess.org. As I was told by gbtami, adding variant Atomic to that wonderful new site requires cooperation by your engine. In Atomic white enjoys a big opening advantage ( some argue, a winning advantage ), therefore there is a great demand for Atomic-960. Many players would join if Atomic / Atomic-960 could be played at that site. Please add variant Atomic to your engine.

ianfab:
Thanks for your feedback.

960 support more or less comes out of the box, so that would not be an issue, however, atomic in general does not fit very well into the design of this fork, which is why I ignored it so far (also see the bottom of https://github.com/ianfab/Fairy-Stockfish/issues/1#issue-335107310 for limitations considered to be difficult to resolve). Atomic still might be feasible, but some problems need to be solved:
* Touching kings are immune from check
  * [ ] legal move validation
  * [ ] evasion and non-evasion move generation
  * [ ] computation of checking pieces/moves
* Doing and undoing of atomic captures
  * [ ] Store captured pieces in state info
  * [ ] How to deal with captured promoted pieces (e.g. for atomic-zh)? Also maintain unpromoted state of captured pieces?
  * [ ] discovered checks (and illegal capture of king blockers) by atomic captures

smartchessguiapp:
If nothing else there is always this open source fork:

https://github.com/ddugovic/Stockfish

Atomic in the above fork became quite reliable over time. I don't know if the code can be injected as is, but at least the ideas used in it can be studied.

I only understand ( somewhat ) GUI programming, so I cannot give any practical advice better than this general one.

ianfab:
As a co-author I know that fork quite well, and my work on it actually was the main motivation to start this fork, since its design and goals impose some limitations that can not be overcome without starting a new fork from scratch. However, the huge amount of variants in this fork make it necessary to implement variants/rules with as few and simple code as possible, otherwise the whole project gets unmaintainable. Therefore, and due to allowing almost arbitrary combination of supported rules, a very different way of implementing the atomic rules is required here.

smartchessguiapp:
May be it is easier then to somehow allow more than one engine in the pychess project than to modify this engine. This was my suggestion in the first place towards gbtami, but he redirected me here, to see if I can succeed with an issue.

[...]

PR #200, opened 2020-10-19
Upgraded ffish.js to 0.4.4 and added validate_fen()

QueensGambit:
Hello @ianfab ,
I implemented an FEN validation method for **apiutil.h** as requested in #87.
+ FEN_VALIDATION validate_fen(const std::string& fen, const Variant* v)

The validation method returns either `FEN_OK` or one of **13** unique error codes.

```C++
enum FEN_VALIDATION : int {
  FEN_MISSING_SPACE_DELIM = -12,
  FEN_INVALID_NB_PARTS = -11,
  FEN_INVALID_CHAR = -10,
  FEN_FILE_SIZE_MISMATCH = -9,
  FEN_INVALID_NUMBER_OF_RANKS = -8,
  FEN_INVALID_POCKET_INFO = -7,
  FEN_INVALID_SIDE_TO_MOVE = -6,
  FEN_INVALID_CASTLING_INFO = -5,
  FEN_INVALID_EN_PASSANT_SQ = -4,
  FEN_INVALID_CHECK_COUNTER = -3,
  FEN_INVALID_HALF_MOVE_COUNTER = -2,
  FEN_INVALID_MOVE_COUTNER = -1,
  FEN_EMPTY = 0,
  FEN_OK = 1
};
```

Each starting fen is now validated in **test.js** including custom variants from **variants.ini**.
In the process of this, I found that `[]` was missing in the starting fen of _makrukhouse_, _xiangqihouse_, _janggihouse_ and _shogun_.
There is a test for each error code in **test.js**.

The following features were added to **ffish.js**:
+ **ffish.validate_fen(fen)** it validates a given chess fen and returns +1 if fen is valid. Otherwise an error code will be returned.
+ **ffish.validate_fen(fen, uciVariant)** it validates a given fen and returns +1 if fen is valid. Otherwise an error code will be returned.
+ **ffish.startingFen(uciVariant)** it returns the starting fen for the given uci-variant.
+ **board.pocket(turn)** it returns the pocket for the given player as a string with no delimeter. All pieces are returned in lower case.
+ uddated **ffish.js** and **ffish-es6** to **0.4.3**.

**Refactoring**
+ replaced logic of `bool first = true` by `save_pop_back(string& s)` to reduce code size.
+ replaced `s += " " + s2;` by `s += " "; s += s2;` to improve speed.

updated **README.md**
+ simplified memory specification
+ added `-std=c++17` to allow building with recent _NNUE_ code


gbtami:
@QueensGambit from one year experience on pychess site the most common cases of user defined invalid FENs are where rook or king positions contradict to castling rights or kings are touching or missing. See https://github.com/gbtami/pychess-variants/blob/master/client/chess.ts#L644
Can you add these additional checks to ffish.js please?

ianfab:
Thanks a lot! It will take me some time to think through all of the possible cases of FENs, so I will perhaps add CR comments step by step and not at once, so maybe you wait a bit with addressing them to save you time. A few initial remarks:
* Some of the startFENs were "intentionally" wrong, since I knew the parser supports those cases, but it makes sense to consolidate those now that we have a validation. The question is if the validator should only accept **the** correct FEN or also others that are accepted by the parser, but perhaps the former makes more sense in order to encourage sticking to standards while still being robust in cases where the validation is not used prior to inputting an FEN, e.g., for FEN differences between cutechess and XBoard, since XBoard e.g. skips en passant and castling fields for variants where they are not applicable. The latter example might also be relevant for checking the startFens, because I think I remember using XBoard style FENs sometimes.
* I could swear I added `-std=c++17` yesterday, but thanks for fixing that. I did not encounter any issues with ffish.js after the merge and seems you didn't either, so that is reaffirming that the NNUE code did not break anything.
* I will perhaps also add the FEN validation to pyffish and to the variant config validator after merging this, unless you want to add this yourself still in this PR.

src/apiutil.h
```diff
@@ -308,3 +316,208 @@ bool hasInsufficientMaterial(Color c, const Position& pos) {
...
+      std::cerr << "Invalid piece character: '" << c << "'." << std::endl;
+      return FEN_INVALID_CHAR;
+    }
+
+  // check for number of ranks
+  int nbRanks = RANK_1;
+  for (; nbRanks <= v->maxRank; ++nbRanks);
+  // check for number of files
+  int nbFiles = FILE_A;
+  for (; nbFiles <= v->maxFile; ++nbFiles);
```
ianfab, line 384: These for loops do not seem to be doing much. Does anything speak against just setting them to the target values right away?

QueensGambit: You're right, setting them directly is better.
```C++
int nbRanks = v->maxRank - RANK_1 + 1;
int nbFiles = v->maxFile - FILE_A + 1;
```

QueensGambit: > The question is if the validator should only accept the correct FEN or also others that are accepted by the parser, but perhaps the former makes more sense in order to encourage sticking to standards

I also allowed using an additional `/` for specifying the pockets as this is the case on [lichess.org](https://lichess.org/).

QueensGambit:
> @QueensGambit from one year experience on pychess site the most common cases of user defined invalid FENs are where rook or king positions contradict to castling rights or kings are touching or missing. See https://github.com/gbtami/pychess-variants/blob/master/client/chess.ts#L644
> Can you add these additional checks to ffish.js please?

Regarding the additional checks.
There are a few variants where these checks need to be skipped:
* **giveaway_variant**, **antichess_variant**, **codrus_variant**: It is allowed to have multiple kings or no king.
* **atomic_variant**, **giveaway_variant**, **antichess_variant**, **codrus_variant**: It is allowed that kings touch.
* **chess960 (castling)**: Checking if the kings and rooks are still on starting squares is not possible because the starting fen is not directly available. However, you could check if the kings are between their rooks.

The implementation of the **atomic_variant** is still missing (https://github.com/ianfab/Fairy-Stockfish/issues/81) but we need to keep this in mind for later.

gbtami:
@QueensGambit correct :)

src/apiutil.h
ianfab: Now that we talk about it, getting the number of ranks/files for a variant is actually a quite common operation in the code base, and I usually for simplicity just write `v->maxRank + 1`, but subtracting `RANK_1` is of course more correct conceptually. Introducing methods for rank and file count in `Variant` to replace those "count = max +1" usages perhaps makes sense, but you do not need to do it as part of the PR, since it is not really related, the code as you wrote it as in your comment should be fine.

ianfab:
In Fairy-SF it is impossible to have more than one (royal) king per side, the code does not support that scenario. Variants with non-royal kings use the commoner piece type, which is strictly separated in the code. So as long as validation code for the king is skipped if none is present (for the respective side), variants like giveaway do not affect such conditions. Regarding the touching kings it is a bit more tricky, because one needs to consider that in Fairy-SF the `KING` piece type is just a placeholder for a royal piece type. The actual movement is determined by the `kingType`, which therefore needs to be considered when checking for "touching" (e.g., consider knightmate) kings.

src/apiutil.h
```diff
@@ -308,3 +316,208 @@ bool hasInsufficientMaterial(Color c, const Position& pos) {
...
+      if (!isdigit(secondChar)) {
+        std::cerr << "Invalid en-passant square. Expects 2nd character to be a non-digit." << std::endl;
+        return FEN_INVALID_EN_PASSANT_SQ;
+      }
+    }
+  }
+
+  // 5) Part
+  // check checkCounting (the karouk_variant doesn't feature a checkCounter within the FEN because the first check wins)
+  if (v->checkCounting && !v->cambodianMoves)
```
ianfab, line 498: Whether the checkCounting field is skipped is not related to cambodian chess, but only to whether the counts are exactly 1. This is a bit ugly from the validation perspective, but it basically means that skipping this field is always allowed, it just implies that the counts are exactly 1. I would also be fine with allowing skipping only for cambodian chess (since allowing it is kind of arbitrary anyway and just related to GUI implementations), but the scenario of cambodian chess with a count > 1 should be supported, i.e., the field should still be allowed to be present as otherwise there are hypothetical variants where no FEN would be valid.

src/apiutil.h
```diff
@@ -308,3 +316,208 @@ bool hasInsufficientMaterial(Color c, const Position& pos) {
...
+      if (c != '-')
+        if (c != 'k' && c != 'K' && c != 'Q' && c != 'q') {
+          std::cerr << "Invalid castling specification: '" << c << "'." << std::endl;
+          return FEN_INVALID_CASTLING_INFO;
+        }
+  }
+
+  // 4) Part
+  // check en-passant square
+  if (v->castling) {
```
ianfab, line 477: This should refer to en passant.

src/apiutil.h
```diff
@@ -308,3 +316,208 @@ bool hasInsufficientMaterial(Color c, const Position& pos) {
...
+
+  int nbProcessedRanks = 1;  // start with 1 and increment after delimeter
+  int curRankWidth = 0;
+  char prevChar = '?';
+  for (char c : fenParts[0]) {
+    if (isdigit(c)) {
+      curRankWidth += c - '0';
+      // if previous digit exists we have a dual digit
+      if (isdigit(prevChar))
+        curRankWidth += 9;
```
ianfab, line 394: This is a bit hacky to me, I needed to read it several times to understand that the additional digit is first counted as 1 and therefore only 9 is added in the next iteration. If it can not easily be solved otherwise, then I think at least writing it as `10 * (prevChar - '0') - 1` makes it clearer, and as an aside also fixes more artifical cases like 01 or 20.

src/apiutil.h
QueensGambit: Is there a boolean to check if the variant has an en passant square?

src/apiutil.h
QueensGambit: I agree that this formulation is a bit hard to understand.
Using `9 * (prevChar - '0')` allows board sizes up to **99x99**.
```C++
curRankWidth += 9 * (prevChar - '0');
```
Alternative, we could not iterate over single characters and convert the number using **stoi()**.

src/apiutil.h
ianfab: Good point 🙂 . I was not really aware that there is no config to disable en passant explicitly without disabling double steps. However, `doubleStep` is a strict requirement for en passant to occur, so that could be used.

src/apiutil.h
```diff
@@ -308,3 +316,208 @@ bool hasInsufficientMaterial(Color c, const Position& pos) {
...
+    std::cerr << "Invalid side to move specification: '" << fenParts[1][0] << "'." << std::endl;
+    return FEN_INVALID_SIDE_TO_MOVE;
+  }
+
+  // 3) Part
+  // check castling rights (exclude seirawan_variant because it has different castling specification)
+  if (v->castling && !v->seirawanGating) {
+    for (char c : fenParts[2])
+      if (c != '-')
+        if (c != 'k' && c != 'K' && c != 'Q' && c != 'q') {
```
ianfab, line 469: Chess960 needs to be considered here, since in shuffling/960 variants the castling rights can be any letters that denote valid files. Since we can not know here if `UCI_Chess960` is enabled, one perhaps always has to assume it might be the case.

src/apiutil.h
```diff
@@ -308,3 +316,208 @@ bool hasInsufficientMaterial(Color c, const Position& pos) {
...
+  // check for number of parts
+  if (fenParts.size() != nbFenParts) {
+    std::cerr << "Invalid number of fen parts. Expected: " << nbFenParts << " Actual: " << fenParts.size() << std::endl;
+    return FEN_INVALID_NB_PARTS;
+  }
+
+  // 1) Part
+  // check for valid characters
+  for (char c : fenParts[0])
+    if (!isdigit(c) && v->startFen.find(c) == std::string::npos) {
```
ianfab, line 374: I think this can break in some cases:
* `+` and `~` signs for promoted pieces that are not present in the starting position
* pieces that can only be obtained via promotion (e.g., a black queen in orda)

If you have not seen it already, I have similar logic in the variant config parser (to validate the start FEN), but it directly checks the array of piece types, which should be safer than relying on the start FEN. https://github.com/ianfab/Fairy-Stockfish/blob/369c55f9f2dce34b9cb65ba78ef1649543d11f59/src/parser.cpp#L336-L347

ianfab:
Apart from the comments I wrote it mostly looks good to me. Ensuring that an FEN is strictly valid is very challenging, so from my perspective it is fine to leave it at the checks you included so far, and then extend it later. In order to have a good understanding of the limitations, I would however like to summarize what is not validated yet:
* Promoted pieces (i.e., does the piece have a promoted form?)
* S-Chess gates
* Cambodian "gates"
* Consistency of castling rights and king/rook positions
* Legality of position (king of side to move in check, etc.)

src/apiutil.h
QueensGambit: I think we can check for `UCI_Chess960` using `v->chess960`.
Therefore, we can relax the castling specification in this case to allow any non-digit character.

ianfab: No, this unfortunately is not possible. The "chess960" attribute in a variant only enforces chess960, even if the UCI option is disabled, but for variants where it is false the UCI option could still be enabled. This logic is to ensure compatibility with the normal UCI option and at the same time allowing to configure variants like caparandom where the UCI option should be ignored. Also see
https://github.com/ianfab/Fairy-Stockfish/blob/369c55f9f2dce34b9cb65ba78ef1649543d11f59/src/position.cpp#L457

QueensGambit:
> [...] Ensuring that an FEN is strictly valid is very challenging, so from my perspective it is fine to leave it at the checks you included so far, and then extend it later, therefore my comments mostly focus on avoiding false negatives.

I concur. Aiming for low false negative rate is more important than reducing false positives here.
This method will most likely mainly be used for validating user inputs.
The variants also differ in popularity making some checks more important than others.
Many additional checks could be added later.

I'll revise the code based on your review.

src/apiutil.h
QueensGambit: I'll include `+`, `~` and `/` as valid characters and will use `v->pieceToChar` insteda.

src/apiutil.h
QueensGambit: I'll relax the castling specification condition to use `isalpha(c)` instead.

src/apiutil.h
QueensGambit: I'll remove this check then to allow skipping the `checkCounter` in cases where only check is required.

src/apiutil.h
QueensGambit: Using `doubleStep` as an identifier instead, runs into the problem for the **breakthrough_variant**.
Here the fen is defined as `pppppppp/pppppppp/8/8/8/8/PPPPPPPP/PPPPPPPP w 0 1`, `firstRankDoubleSteps=false` and `doubleStep=true`.
Is this a bug and should `doubleStep` be set to `false` instead?

ianfab: Setting those values in the absence of pawns does not have any effect, therefore this value might be sloppily defined for some variants. However, in the specific case of breakthrough this really looks like a typo (or rather wrong auto-completion), because if I was too lazy to add doubleStep I would not have added the other one for sure... If we want to fully cover the conditions we could in addition to doubleStep check that pawns are in the set of pieceTypes. I leave it up to you if you prefer to make the condition more specific or make the variant definitions more consistent (or both).

QueensGambit:
I revised the code and added additional checks for _castling_, _touching kings_ and a check for _number of royal kings_ as requested by @gbtami .

The additional castling checks required a lot of more code. Therefore I moved a lot of checks into separate functions to improve maintainability and readability.

All starting fens of all variants returned `FEN_OK`.
Despite, adding new test cases for possible failures, I ensured that all game positions of the example games returned `FEN_OK`.

ianfab:
Thanks! To be honest I am not entirely happy about how some conditions are defined, because they seem to filter down the list of existing variants to the ones where it should apply rather than constructively defining the conditions from the bottom up based on the rules that are prerequisites. However, I can imagine that this perhaps is incredibly hard to do without having too much experience with the "rule space" of this engine (I should perhaps also improve documentation in that regard), and I do not want to unnecessarily hinder the merge of this feature, especially considering that it seems to work well for all existing variants. So if there are no objections I would then at a later point try to modify the conditions to a more bottom up approach, but merge this first version as is.

QueensGambit:
Sounds good to me.
Feel free to change the conditions, as you are much more familiar with the code base and all the different chess variants.

PR #201, opened 2020-10-23
add grasshopper piece

yoav-rozin:
Hey,
I have tried to add a new piece: grasshopper. I created a new movement type (because those who exists are unfit to the grasshopper), I copied the mechanism of the hopper type and I made appropriate changes.
I did few tests: 
- play a game against it [pass]
- ran simulation of a game [pass]
- ran and solve few chess problems with grasshoppers [pass]

TODO: need to add the grasshopper value.
Thanks.


ianfab:
Thanks for your PR. At first sight it mostly looks good to me (did not have time to test yet), I just have a few requests for minor changes so far:
* Please add a piece value and update the bench number at the end of the commit message (e.g. `bench: 1234567`) in order to make CI tests pass
* It would be nice to also add perft tests for grasshopper in perft.sh to cover it in CI
* As soon as CI tests pass, it would be good to either measure speed difference locally or run a regression test on http://www.variantfishtest.org:6543/tests for standard chess in order to check what the impact is on performance.
* Is there a specific reason why you introduced a new enum type and _attacks_bb function for the non-sliding hoppers? If it is because of the namings, I have to say that "rider" was just a bad naming by me, but I failed to come up with something better. In these contexts it more generally refers to movement, move generation, or magic types, so they could in principle also comprise the newly introduced ones, which would slightly simplify the code. So if there is no other reason for the separation, then I would suggest to merge them, and optionally improve the naming if that helps to avoid confusion.

ianfab:
If anything is unclear with my above comments or you need help implementing them, just let me know.

Belzedar94:
Would love to see this merged at some point! It's 90% done but the author seems inactive

ianfab:
Thanks for your contribution. I am closing this now in favor of an updated implementation in #302.

PR #214, opened 2020-12-10
Increase midgame value of aiwok

TomFryers:
Since aiwok = chancellor + fers, the aiwok’s value should be greater than that of the chancellor.

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -341,7 +341,7 @@ enum Value : int {
   AlfilValueMg             = 350,   AlfilValueEg             = 330,
   FersAlfilValueMg         = 700,   FersAlfilValueEg         = 650,
   SilverValueMg            = 630,   SilverValueEg            = 630,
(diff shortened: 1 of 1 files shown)


ianfab:
Thanks for your suggestion. In theory I agree that an Ai-Wok of course can not be worth less than a chancellor. However, in the end what counts is the actual playing strength in relevant variants, therefore it would be nice if you could submit a test for your patch on http://www.variantfishtest.org:6543/tests in order to verify if this improves playing strength for the Ai-Wok variant.

TomFryers:
Submitted (hopefully correctly!).

ianfab:
Thanks, test settings look correct, so let's see how the test will turn out.

ianfab:
The test on fishtest http://www.variantfishtest.org:6543/tests/view/5fd2a7f26e23db221d9e94d9 unfortunately did not show an improvement, but you are very welcome to submit more tests if you have some ideas for improvement.

PR #236, opened 2021-01-07
Add workflow for ffishjs

ianfab:
- Add Makefile for easier compilation
- Run js unit tests as part of CI

### DIFF
--- a/.github/workflows/ffishjs.yml
+++ b/.github/workflows/ffishjs.yml
@@ -1,0 +1,45 @@
+name: ffishjs
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  EM_VERSION: 1.39.16
+  EM_CACHE_FOLDER: 'emsdk-cache'
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      matrix:
+        node-version: [12.x]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup cache
+        id: cache-system-libraries
+        uses: actions/cache@v2
+        with:
+          path: ${{env.EM_CACHE_FOLDER}}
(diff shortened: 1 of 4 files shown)


ianfab:
@QueensGambit Please let me know if you have any feedback. For now I kept it simple just to quickly have at least basic CI for ffishjs, but I guess it makes sense to add some makefile flags for e.g. largeboards, es6, and debug later on to be able to use the makefile in general and not just for CI purposes.

QueensGambit:
Thank you @ianfab , this looks good to me.
I only have a few comments and suggestions.
The new Makefile builds the release version of **ffish.js** as I publish it to npm.
The flag `-DLARGEBOARDS`, which has already been set in the Makefile, is required because some unit tests are done on large boards.

For the unit tests, it could be beneficial to change:
`-s ASSERTIONS=0 -s SAFE_HEAP=0`
into
`-s ASSERTIONS=1 -s SAFE_HEAP=1`
instead. This reduces performance but enables more checks.

We also probably do not need `-O3` as optimization level but could use e.g. `-O2` or `-O1` instead to speed up compilation time.
As far as I remember, if a too low optimization level has been set and all debug options are active, you may encounter a timeout for the first Unit test by [**Mocha**](https://mochajs.org/). This is because it can then take a few seconds for the web assembly file to load. We could [increase the default timeout](https://stackoverflow.com/questions/23492043/change-default-timeout-for-mocha) of 2000ms to e.g. 5000ms to avoid this by calling mocha with the command line argument `--timeout 5000` in [package.json](https://github.com/ianfab/Fairy-Stockfish/blob/master/tests/js/package.json).

ianfab:
@QueensGambit Thanks, I extended the makefile now to have additional compilations options and simplified the readme accordingly. CI now uses debug=yes and optimize=no. All makefile flags default to the release version (i.e., also largeboards=yes).

QueensGambit:
Great! This makes the building process easier and lets us know if new changes might break the **ffish.js** library.

PR #250, opened 2021-02-01
Help request: Trying to implement dabbaba piece

prendradjaja:
Hello! Thank you for this project of yours :) If this should be an issue instead of a pull request, apologies -- let me know!

I'm trying to implement the dabbaba, which is similar to the fers, alfil, wazir that you already have: (An unrestricted dabbaba has four legal moves: 2 spaces north, 2 spaces west, 2 spaces east, and 2 spaces south. These moves are jumps, like the knight or the alfil.)

Diagram showing the dabbaba's movement as described above

I've looked to see how existing pieces are implemented and implementing a simple new one like this looks relatively straightforward -- however, I've definitely done something wrong -- see [log below](https://github.com/ianfab/Fairy-Stockfish/pull/250#discussion_r567514991) showing:
- `./stockfish check variants.ini` reports no problems
- The (new) dabbaba (D) can be displayed
- I can move the (existing) alfil (A) from g1 to e3
- I cannot move the (new) dabbaba (D) from b1 to b3

I'm using a Mac and compiling with `make build`.

If you'd be able to help me see what I might be doing wrong, I'd definitely appreciate it -- thanks! (I added some PR comments describing what I currently understand about how pieces are implemented, in case there's something wrong there that might be helpful.)


src/piece.cpp
```diff
@@ -110,6 +110,14 @@ namespace {
       p->stepsCapture = {2 * SOUTH_WEST, 2 * SOUTH_EAST, 2 * NORTH_WEST, 2 * NORTH_EAST};
```
prendradjaja, line 110: (Unrelated to this line, just making a "thread")

Log showing successful movement of (existing) alfil but failure to move (new) dabbaba, as described in PR description above

```
$ ./stockfish check variants.ini 
Fairy-Stockfish 310121 by Fabian Fichter
Parsing variant: 3check-crazyhouse
Parsing variant: coffeehouse
Parsing variant: makrukhouse
Variant template 'xiangqi' does not exist.
Variant template 'janggi' does not exist.
Parsing variant: anti-losalamos
Parsing variant: indiangreat
maxRank - Invalid value 10 for type Rank
maxFile - Invalid value 10 for type File
promotionRank - Invalid value 10 for type Rank
Parsing variant: upsidedown
Parsing variant: peasant
Parsing variant: weak
Parsing variant: semitorpedo
Variant template 'janus' does not exist.
Parsing variant: pawnsonly
Parsing variant: tictactoe
Parsing variant: cfour
Parsing variant: flipersi
Parsing variant: flipello
Variant template 'grand' does not exist.
Parsing variant: shogun
Parsing variant: orda
Parsing variant: myvariant

$ ./stockfish load variants.ini 
Fairy-Stockfish 310121 by Fabian Fichter
Variant template 'xiangqi' does not exist.
Variant template 'janggi' does not exist.
Variant template 'janus' does not exist.
Variant template 'grand' does not exist.
xboard
variant myvariant
setup (-) 8x8+0_fairy rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RDBQKBAR w KQkq - 0 1
piece P& fmWfceFifmnD
piece N& N
piece B& B
piece R& R
piece Q& RB
piece A& A
piece K& KO2
piece D& D
d

 +---+---+---+---+---+---+---+---+
 | r | n | b | q | k | b | n | r |8  
 +---+---+---+---+---+---+---+---+
 | p | p | p | p | p | p | p | p |7
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |6
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |5
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |4
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |3
 +---+---+---+---+---+---+---+---+
 | P | P | P | P | P | P | P | P |2
 +---+---+---+---+---+---+---+---+
 | R | D | B | Q | K | B | A | R |1 *
 +---+---+---+---+---+---+---+---+
   a   b   c   d   e   f   g   h

Fen: rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RDBQKBAR w KQkq - 0 1
Sfen: rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RDBQKBAR b - 1
Key: 5E9769DF789A7BE3
Checkers: 
g1e3
d

 +---+---+---+---+---+---+---+---+
 | r | n | b | q | k | b | n | r |8 *
 +---+---+---+---+---+---+---+---+
 | p | p | p | p | p | p | p | p |7
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |6
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |5
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |4
 +---+---+---+---+---+---+---+---+
 |   |   |   |   | A |   |   |   |3
 +---+---+---+---+---+---+---+---+
 | P | P | P | P | P | P | P | P |2
 +---+---+---+---+---+---+---+---+
 | R | D | B | Q | K | B |   | R |1  
 +---+---+---+---+---+---+---+---+
   a   b   c   d   e   f   g   h

Fen: rnbqkbnr/pppppppp/8/8/8/4A3/PPPPPPPP/RDBQKB1R b KQkq - 1 1
Sfen: rnbqkbnr/pppppppp/8/8/8/4A3/PPPPPPPP/RDBQKB1R w - 2
Key: 6450E4DA00EC5D64
Checkers: 
h7h6
b1b3
Error (unknown command): b1b3
d

 +---+---+---+---+---+---+---+---+
 | r | n | b | q | k | b | n | r |8  
 +---+---+---+---+---+---+---+---+
 | p | p | p | p | p | p | p |   |7
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   | p |6
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |5
 +---+---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |   |4
 +---+---+---+---+---+---+---+---+
 |   |   |   |   | A |   |   |   |3
 +---+---+---+---+---+---+---+---+
 | P | P | P | P | P | P | P | P |2
 +---+---+---+---+---+---+---+---+
 | R | D | B | Q | K | B |   | R |1 *
 +---+---+---+---+---+---+---+---+
   a   b   c   d   e   f   g   h

Fen: rnbqkbnr/ppppppp1/7p/8/8/4A3/PPPPPPPP/RDBQKB1R w KQkq - 0 2
Sfen: rnbqkbnr/ppppppp1/7p/8/8/4A3/PPPPPPPP/RDBQKB1R b - 3
Key: 920736C3D788CF2B
Checkers: 
^C
```

src/piece.cpp
```diff
@@ -110,6 +110,14 @@ namespace {
...
       return p;
   }
+  PieceInfo* dabbaba_piece() {
```
prendradjaja, line 113: I used `wazir_piece` as a template for this.

src/piece.cpp
```diff
@@ -110,6 +110,14 @@ namespace {
...
+      PieceInfo* p = new PieceInfo();
+      p->name = "dabbaba";
+      p->betza = "D";
```
prendradjaja, line 116: I'm guessing the Betza notation isn't important for the use case I'm showing (xboard via terminal), which is indeed all I need -- added this anyway, but not sure if "D" needs to be defined elsewhere in Fairy-Stockfish.

src/piece.cpp
```diff
@@ -110,6 +110,14 @@ namespace {
       p->stepsCapture = {2 * SOUTH_WEST, 2 * SOUTH_EAST, 2 * NORTH_WEST, 2 * NORTH_EAST};
       return p;
   }
+  PieceInfo* dabbaba_piece() {
+      PieceInfo* p = new PieceInfo();
+      p->name = "dabbaba";
```
prendradjaja, line 115: I assume this piece name is just used so that `variants.ini` (and maybe elsewhere) can refer to the piece name, so this just needs to match my `variants.ini` change (which it does)

src/piece.cpp
```diff
@@ -110,6 +110,14 @@ namespace {
...
+      p->betza = "D";
+      p->stepsQuiet = {2 * SOUTH, 2 * WEST, 2 * EAST, 2 * NORTH};
+      p->stepsCapture = {2 * SOUTH, 2 * WEST, 2 * EAST, 2 * NORTH};
```
prendradjaja, line 118: I think:

- "quiet" = not capturing
- "capture" = capturing

- "steps" = jumps like a knight or alfil
- "slider" = sliding like a rook or bishop
- "hopper" = jump-and-capture like a checkers piece, but  I'm not sure what `hopperQuiet` would mean

src/piece.cpp
prendradjaja: I also tried to move the dabbaba diagonally and only one square -- left out of the above log for simplicity.

ianfab:
Hi. In principle the definition of the new piece type looks fine. However, due to the special handling of the king, it has to be the last piece type (in the enum), so any other new piece type needs to be inserted somewhere between FAIRY_PIECES and KING. This should fix the issue with the move generation.

Another issue is that new piece types also require to add a piece value in types.h to the PieceValue array. However, this is only relevant for evaluation, the move generation itself should already work even without this change, although a static assert might be triggered during compilation that tries to prevent this oversight.

src/piece.cpp
ianfab: Yes, your understanding is correct. Hopper generally refers to pieces that need a screen/hurdle to jump, but does not necessarily imply a capture move, although the cannon might be the most well known example. So `hopperQuiet` refers to a cannon-like move where the target square is empty, e.g., like for the cannon in Janggi.

prendradjaja:
> Hi. In principle the definition of the new piece type looks fine. However, due to the special handling of the king it has to be the last piece type (in the enum), so any other new piece type needs to be inserted somewhere between QUEEN and KING. This should fix the issue with the move generation.

Ah, thank you so much!

> Another issue is that new piece types also require to add a piece value in types.h to the PieceValue array. However, this is only relevant for evaluation, the move generation itself should already work even without this change, although a static assert might be triggered during compilation that tries to prevent this oversight.

Gotcha! Makes sense.

PR #251, opened 2021-02-02
Updated ffish.js to 0.6.0, fixes ffish.validateFen(), ffish.readGamePGN(pgn), added board.variant(), atomic support

QueensGambit:
Hi @ianfab ,

I noticed that you integrated atomic into Fairy-Stockfish (https://github.com/ianfab/Fairy-Stockfish/pull/238 ).
Many people will enjoy using it in **ffish.js@0.6.0**, too.

## Bug fixes
* **ffish.readGamePGN(pgn)** sometimes pushed  `{` as SanMoves. This is now fixed by incrementing `curIdx`.
*  **ffish.validateFen()** is more robust now and allows the placeholder `-` for variants which do not have a no-progress counter or castling rights. The unit tests have been extended accordingly to also check the fen after a board object has been created using the starting-fen.

## New features

* **board.variant()** it returns the uci-variant of the board.

## Miscellaneous

* **board.push()**, **board.pushSan()**, **board.sanMove()**, **board.variationSan()** now inform the user via _stderr_ if a given move is invalid and will return **false** or **""**. I noticed that `UCI::to_move(const Position& pos, string& str)` already does a legal move check. Therefore, this additional runtime overhead is very small.

### DIFF
--- a/src/apiutil.h
+++ b/src/apiutil.h
@@ -692,6 +692,19 @@ bool no_king_piece_in_pockets(const std::array<std::string, 2>& pockets) {
     return pockets[WHITE].find('k') == std::string::npos && pockets[BLACK].find('k') == std::string::npos;
 }
 
+Validation check_digit_field(const std::string& field)
+{
+    if (field.size() == 1 && field[0] == '-') {
+        return OK;
+    }
+    for (char c : field)
+        if (!isdigit(c)) {
+            return NOK;
+        }
+    return OK;
+}
+
+
 FenValidation validate_fen(const std::string& fen, const Variant* v) {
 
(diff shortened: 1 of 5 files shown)


ianfab:
Thanks for the update.

Please note that, as mentioned in the PR you linked, the atomic variant is not fully compatible to the rules used on lichess yet, since check(-mate) and stalemate are ignored so far and only king explosions end the game. In game play of the engine the impact is very limited, since with optimal play this usually just prolongs PV for checkmate lines by 1-2 ply (but the GUI will adjudicate earlier), but for use cases of ffishjs this might be more problematic. I plan to mitigate/resolve the remaining rule differences, but it will take time.

PR #286, opened 2021-04-08
Refactor FEN validation

ianfab:
- Support X-FEN validation
  - Support fetching variant list in pyffish
  - Use variant configuration properties more consistently
  - Fix a few corner cases
  - Improve test coverage
  - More consistent formatting

Still missing:
- Validation of S-Chess and Ouk gating ("castling") info

### DIFF
--- a/src/apiutil.h
+++ b/src/apiutil.h
@@ -26,10 +26,6 @@
 #include "variant.h"
 
 
-namespace PSQT {
-void init(const Variant* v);
-}
-
 enum Notation {
     NOTATION_DEFAULT,
     // https://en.wikipedia.org/wiki/Algebraic_notation_(chess)
@@ -50,6 +46,8 @@ Notation default_notation(const Variant* v) {
         return NOTATION_SHOGI_HODGES_NUMBER;
     return NOTATION_SAN;
 }
+
+namespace SAN {
 
 enum Disambiguation {
     NO_DISAMBIGUATION,
@@ -86,7 +84,8 @@ std::string piece(const Position& pos, Move m, Notation n) {
 }
 
 std::string file(const Position& pos, Square s, Notation n) {
-    switch (n) {
+    switch (n)
+    {
     case NOTATION_SHOGI_HOSKING:
     case NOTATION_SHOGI_HODGES:
     case NOTATION_SHOGI_HODGES_NUMBER:
@@ -101,7 +100,8 @@ std::string file(const Position& pos, Square s, Notation n) {
 }
 
 std::string rank(const Position& pos, Square s, Notation n) {
-    switch (n) {
+    switch (n)
+    {
     case NOTATION_SHOGI_HOSKING:
     case NOTATION_SHOGI_HODGES_NUMBER:
         return std::to_string(pos.max_rank() - rank_of(s) + 1);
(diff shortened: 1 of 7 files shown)


ianfab:
@QueensGambit @gbtami If you have any input on the scenarios that should be covered for FEN validation please let me know.

QueensGambit:
One convenient feature would be to allow a missing 50-move rule and full move counter.
In that case, we assume a 50-move rule counter of 0 and a full move counter as 1.
This is also possible on lichess.org.

e.g. `4b1k1/1p3p2/4pPp1/p2pP1P1/P2P4/1P1B4/8/2K5 w - -`
and `4b1k1/1p3p2/4pPp1/p2pP1P1/P2P4/1P1B4/8/2K5 w - - 0`
will be interpreted as a valid FEN and result in 
`4b1k1/1p3p2/4pPp1/p2pP1P1/P2P4/1P1B4/8/2K5 w - - 0 1`.

This behavior would also require some changes to the `FEN_INVALID_NB_PARTS` check.

Additionally, if the side to move information is missing as well, we could assume that it is White's turn to move, again same as on lichess.org.

e.g. `4b1k1/1p3p2/4pPp1/p2pP1P1/P2P4/1P1B4/8/2K5`
will be interpreted as a valid FEN and result in 
`4b1k1/1p3p2/4pPp1/p2pP1P1/P2P4/1P1B4/8/2K5 w - - 0 1`.

ianfab:
Thanks, I applied some changes to allow skipping move counts, castling and en passant, or all of them including side to move. Side to move now defaults to white, since that seems more intuitive. The tests now also cover these scenarios.

QueensGambit:
A Janggi player,  called @cjssh1002 , reported a few issues in [LiGround#172](https://github.com/ml-research/liground/issues/172) about ffish.js regarding the FEN notation.
In the future I can also ping, if anything **ffish.js** / _FairyStockfish_ related is reported on _LiGround_.


> 2.janggi/janggicasual/janggitraditional/janggimodern
> There are 4 types of rules. We hope to support all four types of rules.
> 
>     1. When running janggi on Winboard/xBoard, use the CECP protocol. Currently, Riground uses the UCI protocol. (Add support for moving coordinates)
>        So the starting rank is different.
>        UCI starts at rank 1, but CECP starts at rank 0.
>        So the PGN is not compatible.
>        Separate the coordinates or ask for support to run in CECP mode.
> 
>     2. FEN compatibility recognition problem [ex. v->add_piece(JANGGI_ELEPHANT,'b','e');]
>        rhea1aehr/4k4/1c5c1/p1p1p1p1p/9/9/P1P1P1P1P/1C5C1/4K4/RHEA1AEHR w 0 1 (Winboard/Xboard)
>        rnba1abnr/4k4/1c5c1/p1p1p1p1p/9/9/P1P1P1P1P/1C5C1/4K4/RNBA1ABNR w - - 0 1 (Riground)
> 
> 
> Ask them to be able to recognize the following pieces as well.
> n/N = h/H
> b/B = e/E
> w - - 0 1 = w 0 1

Would changing the piece names for Janggi break certain things, or is a sort of _hack_ required to convert the piece names for Janggi?

ianfab:
The support for the alternative piece symbols is just for compatibility with Winboard/XBoard, e.g., also for Xiangqi. Since Fairy-Stockfish supports parsing such FENs, we can also allow it in the FEN validation. On the output side I would however always use the standard names.

QueensGambit:
Sounds good to me.

PR #310, opened 2021-05-20
Create a workflow build for linux

TheYoBots:
No functional change


ianfab:
Thanks for your PR. However, I do not really see the use case of this. The average user is supposed to download the latest release, and the CI builds for Windows I only included because it can be tedious to compile on Windows. The Linux build would only be useful for the small fraction of users who use Linux, want the latest dev build, but are not able to compile it themselves. Since github workflow runtime and storage is limited, I prefer to use it for more relevant use cases.

PR #317, opened 2021-05-23
Port wider UCI_Elo range from ddugovic/Stockfish (closes #316)

### LINKED ISSUE #316: Wider UCI_Elo range
I am finally in the process of finishing niklasf/fishnet#174, which was pretty painless so far. All the variants are working nicely out of the box.

The only remaining todo appears to be that ddugovic/Stockfish had support for even weaker skill levels. Currently it feels like Fairy Stockfish might be too strong for beginners, even on the lowest available level.

For reference, this UCI_Elo range was used by fishnet:
https://github.com/niklasf/fishnet/blob/78782635b5b533ffda68b6bd1f739e896f21ae44/src/api.rs#L226-L233

And some related commits from ddugovic/Stockfish:
ddugovic/Stockfish@b8efae6ef396ad968c251b85145a6187f3a062fe, ddugovic/Stockfish@0d35a730007c5f1b2607a9333532f6e7a8bb505e

niklasf:
The Skill Level option already allowed values from -20 to 20, but not via UCI_Elo. Expands the UCI_Elo range to linearly correspond to the entire range of levels.

The linear interpolation is almost equivalent to the original Elo -> Skill level mapping:

image

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -391,12 +391,9 @@ void Thread::search() {
 
   // Pick integer skill levels, but non-deterministically round up or down
   // such that the average integer skill corresponds to the input floating point one.
-  // UCI_Elo is converted to a suitable fractional skill level, using anchoring
-  // to CCRL Elo (goldfish 1.13 = 2000) and a fit through Ordo derived Elo
-  // for match (TC 60+0.6) results spanning a wide range of k values.
   PRNG rng(now());
   double floatLevel = Options["UCI_LimitStrength"] ?
-                      std::clamp(std::pow((Options["UCI_Elo"] - 1346.6) / 143.4, 1 / 0.806), 0.0, 20.0) :
+                      std::clamp((Options["UCI_Elo"] - 1500.0) / 75.0, -20.0, 20.0) :
                         double(Options["Skill Level"]);
   int intLevel = int(floatLevel) +
                  ((floatLevel - int(floatLevel)) * 1024 > rng.rand<unsigned>() % 1024  ? 1 : 0);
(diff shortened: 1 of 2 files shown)


niklasf:
On second thought, even though the linear approximation is good in the original range, I am not sure if it's sensible for the expanded range. So definitely feel free to reject this.

ianfab:
Thanks, I will do some testing with cutechess to see how Elo behaves relative to UCI_Elo in the extended region. If you already have such results, please let me know.

ianfab:
Up to around Skill Level -5 a linear formula still seems to work roughly, but below that it does not really reflect the compressed Elo. Here a tournament of the lower skill levels:
```
   # PLAYER    :  RATING  POINTS  PLAYED   (%)
   1 3         :  1817.8   143.5     150    96
   2 2         :  1694.4   137.5     150    92
   3 1         :  1541.4   128.5     150    86
   4 0         :  1350.0   115.5     150    77
   5 -2        :  1234.0   107.0     150    71
   6 -1        :  1213.8   105.5     150    70
   7 -3        :  1024.8    91.5     150    61
   8 -4        :   822.0    76.0     150    51
   9 -5        :   605.9    56.0     150    37
  10 -7        :   559.0    51.0     150    34
  11 -6        :   536.1    48.5     150    32
  12 -8        :   406.4    34.0     150    23
  13 -10       :   401.9    33.5     150    22
  14 -15       :   322.2    25.0     150    17
  15 -12       :   317.2    24.5     150    16
  16 -20       :   297.1    22.5     150    15
```
This also intuitively makes sense, since the centipawn loss should be expected to increase super-linearly. I prepared an alternative formula in #319, currently testing it.

niklasf:
The results make sense to me. Closing in favor of #319.

PR #322, opened 2021-06-15
Musketeer variant

alexobviously:
The base functionality and configuration for musketeer chess is here. It seems to work okay for me in UCI, but I'll be doing some testing with pychess again too. Just leaving this here so we can compare and discuss.


alexobviously:
I'll look into those failures, hold on

ianfab:
Thanks. I think part of the failures are due to the FEN validation code for ffishjs/pyffish (in apiutil.h) which might not be aware of the "*" notation yet. However, there also seems to be a crash in the Travis CI, which would be the more critical issue to investigate.

alexobviously:
> Thanks. I think part of the failures are due to the FEN validation code for ffishjs/pyffish (in apiutil.h) which might not be aware of the "*" notation yet. However, there also seems to be a crash in the Travis CI, which would be the more critical issue to investigate.

Do you have any idea how I can recreate the Travis CI tests locally? Because I'm running the same tests that fail (instrumented.sh) and not getting any errors on my machine.

ianfab:
So I guess you are executing something like `make clean && make -j2 ARCH=x86-64-modern debug=yes optimize=no build > /dev/null && ../tests/instrumented.sh --valgrind`? If this does not fail, then maybe it is due to different compiler versions in Travis CI compared to your local environment, I have not tried it myself yet. By the way, just to avoid confusion, next time you push the tests will be run with github actions instead of Travis CI, but they basically cover the same tests.

alexobviously:
> So I guess you are executing something like `make clean && make -j2 ARCH=x86-64-modern debug=yes optimize=no build > /dev/null && ../tests/instrumented.sh --valgrind`? If this does not fail, then maybe it is due to different compiler versions in Travis CI compared to your local environment, I have not tried it myself yet. By the way, just to avoid confusion, next time you push the tests will be run with github actions instead of Travis CI, but they basically cover the same tests.

ah you replied faster than I thought - I deleted that because I found the mistake right after posting (as is usually the case). I just missed the `debug=yes` flag, so yeah, I can recreate this actually. it's failing here:

```
inline Color color_of(Piece pc) {
  assert(pc != NO_PIECE);
  return Color(pc >> PIECE_TYPE_BITS);
}
```
on the assertion, so I guess at some point I'm checking the color of a piece that doesn't exist, maybe in the gate after being gated or something like that.

musketeerchess:
Hi Alex and Fabian
Any new progress in debugging the Musketeer Merge ?

ianfab:
Sorry, Zied crossed a line in a personal conversation, so this unfortunately won't have any chance of merging. Still thanks for your efforts.

PR #333, opened 2021-07-27
Fix variant initialization for libraries

ianfab:


gbtami:
Tested with local pychess. Works like charm!

ianfab:
Thanks for testing. I wanted to check once more if I did not break 960 because of the removal of the UCI_Chess960 setting (which I did for consistency with UCI_Variant), but as far as I can see we already have 960 covered in the tests, so it should be good.

PR #342, opened 2021-08-06
Fixed blasted pieces incorrectly updating key's castling rights

majormink:
This fixes a bug with how blasting in atomic affects castling rights' keys.

reproduction:
turn off Fast in `pos_is_ok`
run in debug:
```
xboard
variant atomic
setboard 7k/8/8/7q/8/8/7P/4K2R b K - 0 1
usermove h5h2
```
the key will differ when recalculated


ianfab:
Thanks, looks good.

PR #356, opened 2021-09-13
Enable re-usage of Standard Chess networks for Amazon/Almost chess

Belzedar94:


ianfab:
Thanks for the suggestion. While I get your point that chancellors and amazons are somewhat similar in strength to the queen and the starting positions have a one to one replacement of them, these two variants are a very different case than what I did for the NNUE reusage in other variants.

For cambodian, placement, etc. the set of piece types is exactly identical to the respective other variant from which they are reusing the net, so the input layer by definition matches perfectly, while for chancellor and amazon it is more of a coincidence that these pieces use the same NNUE index as the queen (as the pieces are not perfectly ordered by value). Therefore this is a bit hacky to me and I would not like to use this as a default behavior. The decision for such a custom behavior should be made intentionally by the user, e.g., by copying/renaming the net. Therefore I am closing this.

Belzedar94:
I totally agree with your approach. At first sight, I thought it was a similar case to the aliases you already defined, but it's true that for these two particular variants it is not the case at all, so I agree this would be a totally wrong standard. I'll do the hack for both experiments.

Thank you! :)

PR #360, opened 2021-09-14
Autoload variants.ini if specified in by an environment variable

thearst3rd:
Hi, thanks for making Fairy-Stockfish! I've been enjoying using it.

I'm just making this pull request to put this out there if anyone is interested. I've been messing around with custom variants, and think it would be convenient to not need to specify the load command if you keep your variants file in the same place. This PR will have Fairy-Stockfish check the environment variant `FAIRY_STOCKFISH_VARIANT_PATH`, and if it exists, it will load the file at this path as the variant config. This will get overridden by a "load" command on the command line.

I tested it on Windows 10 and Linux Mint 20.2, and it works pretty well!

I have a question though, and maybe it should go in an issue. I noticed that the "load" command, when entered as a command line arg or in the UCI loop, cannot handle paths with spaces in it. So, on Windows if I try loading `C:\Program Files\fairy-stockfish\variants.ini`, it would try and fail to load two files `C:\Program` and `Files\fairy-stockfish\variants.ini`. I'm not sure what we can do about this since users might want to load multiple files in one load command, maybe we could escape strings? Seems to add unneeded complexity ¯\\\_(ツ)\_/¯ Setting the "VariantPath" UCI option directly works though (both in code and in the UCI loop), so that's what I'm doing here.

Thanks for writing Fairy-Stockfish, and thanks for your time! :smile:

### DIFF
--- a/src/uci.cpp
+++ b/src/uci.cpp
@@ -16,6 +16,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include <cstdlib>
 #include <cassert>
 #include <cmath>
 #include <iostream>
@@ -286,6 +287,11 @@ void UCI::loop(int argc, char* argv[]) {
   XBoard::stateMachine = new XBoard::StateMachine(pos, states);
   // UCCI banmoves state
   std::vector<Move> banmoves = {};
+
+  // Check environment for variants.ini file
+  char *envVariantPath = std::getenv("FAIRY_STOCKFISH_VARIANT_PATH");
+  if (envVariantPath != NULL)
+      Options["VariantPath"] = std::string(envVariantPath);
 
   do {
       if (argc == 1 && !getline(cin, cmd)) // Block here waiting for input or EOF
--- a/src/uci.cpp
+++ b/src/uci.cpp
@@ -288,7 +288,12 @@ void UCI::loop(int argc, char* argv[]) {
   // UCCI banmoves state
   std::vector<Move> banmoves = {};
 
-  if (argc == 1)
+  if (argc > 1 && (std::strcmp(argv[1], "noautoload") == 0))
+  {
+      cmd = "";
+      argc = 1;
+  }
+  else if (argc == 1 || !(std::strcmp(argv[1], "load") == 0))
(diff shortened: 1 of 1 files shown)


ianfab:
Thanks for your suggestion! Although I personally did not need this yet, I think it can be useful. However, I would suggest to only skip the loading of the environment variable explicitly when `argv[1]` is `load`, since in other cases it might still be desirable to load the INI, e.g., when you would have something like `./stockfish bench my-configured-variant-from-ini`.

I think you are right about the issue with spaces, since the command line arguments are simply joined together using spaces, so it is impossible to distinguish afterwards without quoting/escaping. Considering how specifying multiple paths is handled for UCI options, the canonical way to solve that perhaps would be to do the iteration over separated paths in the VariantPath option itself (splitting at the `:`/`;` separator), and just pass the entire rest of the line after `load` to the `VariantPath` option without further parsing. This should fix the issue with spaces and at the same time also allow loading multiple INIs via `VariantPath` without using `load`, which would make it more consistent. In any case feel free to raise a bug for this.

thearst3rd:
Yep, makes sense, I can add that! I was also thinking, it might be nice to have a command like `noautoload` which overrides the auto-loading without loading in something new. Obviously that name can change, just what I thought up.

ianfab:
Thanks. Regarding the `noautoload` I am not sure though if it is worth adding, since using the environment variable already is pretty much an expert feature, and skipping loading the environment variable would be even more so, and could e.g. also be achieved by leaving `load` empty, i.e., `./stockfish load`. Actually currently this should already be possible without any code changes. However, I do not have a strong opinion on this, whether to be more explicit or more concise.

thearst3rd:
Hmm, that's a good point actually. I didn't consider that just leaving "load" blank would work without throwing any errors. I already added it in the commit last night but I'm not very attached to it

ianfab:
Ok, let's not overthink it, I will just merge it as is, since the code is fine. If we do not find the `noautoload` useful, we can later remove it, or maybe it might turn more useful if we introduce something similar for NNUE nets or the like.

PR #362, opened 2021-09-17
Expose additional game-end functions in ffish.js

thearst3rd:
In pyffish, there are functions exposed to check for insufficient material, optional draws (50 move, threefold, etc), and to get the resulting value of the game. Such functionality didn't exist in ffish.js, and I wanted it exposed so I could mess around with it in [my toy ffish.js+chessboardx project](https://github.com/thearst3rd/ffish-test) I've been working on lately.

There are some things I'm not sure I'm doing totally right - for example, `is_immediate_game_end` and `is_optional_game_end` have a parameter which can specify the value of the game (if the game did immediately or optionally end). Right now, I'm only returning a boolean of whether or not the game ended. Maybe it would be nice if we could return a javascript array such that the first item is the boolean, and the second item (if the first is true) will be the value? I'm not very familiar with emscripten but I'm hoping something like that will be possible.

I still have some questions, such as it seems that no game end function automatically considers insufficient material (for example, it is totally ignored on pychess.org analysis boards) but I think these functions match up pretty well with the pyffish versions so I'm assuming that's ok.

I've been really enjoying using this library so far and I hope these functions may be useful!

CC: @QueensGambit

### DIFF
--- a/src/ffishjs.cpp
+++ b/src/ffishjs.cpp
@@ -263,6 +263,34 @@ public:
     return pos.game_ply();
   }
 
+  int game_result() const {
+    Value result;
+    bool gameEnd;
+    if (MoveList<LEGAL>(pos).size() > 0)
+      return VALUE_NONE;
+    gameEnd = pos.is_immediate_game_end(result);
+    if (!gameEnd)
+        result = pos.checkers() ? pos.checkmate_value() : pos.stalemate_value();
+    return result;
+  }
+
+  bool is_immediate_game_end() const {
+    return pos.is_immediate_game_end();
+  }
+
+  bool is_optional_game_end() const {
+    return pos.is_optional_game_end();
+  }
+
+  // Need to use different name than has_insufficient_material
+  bool has_insufficient_material_ffish(bool turn) const {
+    return has_insufficient_material(turn ? WHITE : BLACK, pos);
+  }
+
+  bool has_insufficient_material_ffish() const {
+    return has_insufficient_material(WHITE, pos) && has_insufficient_material(BLACK, pos);
+  }
+
   bool is_game_over() const {
     for (const ExtMove& move: MoveList<LEGAL>(pos))
       return false;
@@ -626,6 +654,11 @@ EMSCRIPTEN_BINDINGS(ffish_js) {
     .function("fullmoveNumber", &Board::fullmove_number)
     .function("halfmoveClock", &Board::halfmove_clock)
     .function("gamePly", &Board::game_ply)
+    .function("gameResult", &Board::game_result)
+    .function("isImmediateGameEnd", &Board::is_immediate_game_end)
+    .function("isOptionalGameEnd", &Board::is_optional_game_end)
+    .function("hasInsufficientMaterial", select_overload<bool(bool) const>(&Board::has_insufficient_material_ffish))
+    .function("hasInsufficientMaterial", select_overload<bool() const>(&Board::has_insufficient_material_ffish))
     .function("isGameOver", &Board::is_game_over)
     .function("isCheck", &Board::is_check)
     .function("isBikjang", &Board::is_bikjang)


ianfab:
Thanks! I will be travelling and therefore likely not be very responsive till Monday, and won't be able to review/merge before then, but on a first glance it looks good to me.

src/ffishjs.cpp
```diff
@@ -263,6 +263,34 @@ class Board {
...
+  bool is_immediate_game_end() const {
+    return pos.is_immediate_game_end();
+  }
+
+  bool is_optional_game_end() const {
+    return pos.is_optional_game_end();
+  }
+
+  // Need to use different name than has_insufficient_material
+  bool has_insufficient_material_ffish(bool turn) const {
```
QueensGambit, line 286: Please remove comment  
```C++
// Need to use different name than has_insufficient_material
```
 and remove suffix `_ffish`.
 
Below you can use
 ```C++
 return Stockfish::has_insufficient_material(turn ? WHITE : BLACK, pos);
 ```
 to avoid ambiguity.

src/ffishjs.cpp
```diff
@@ -263,6 +263,34 @@ class Board {
...
+    return has_insufficient_material(turn ? WHITE : BLACK, pos);
+  }
+
+  bool has_insufficient_material_ffish() const {
```
QueensGambit, line 290: Please remove suffix `_ffish`.
 
Below you can use
 ```C++
    return Stockfish::has_insufficient_material(WHITE, pos) && Stockfish::has_insufficient_material(BLACK, pos);
 ```
 to avoid ambiguity.

src/ffishjs.cpp
QueensGambit: Also please rename into `is_insufficient_material()`.

The **ffish.js** library aims to have a syntax similar to python-chess in cases where this is possible.
In python-chess this function is called

```
is_insufficient_material() → bool[source]
    Checks if neither side has sufficient winning material (has_insufficient_material()).
```    
* https://python-chess.readthedocs.io/en/latest/core.html#outcome

If you like, you can also add
```
has_insufficient_material(color: chess.Color) → bool
```
accordingly.

src/ffishjs.cpp
```diff
@@ -263,6 +263,34 @@ class Board {
     return pos.game_ply();
   }
 
+  int game_result() const {
+    Value result;
```
QueensGambit, line 267: The idea of using `Value` here looks good.
However, I'm unsure if this is really meaningful to the user.
For instance `VALUE_KNOWN_WIN` corresponds to `10000` and `VALUE_MATE` corresponds to `32000`.
There is also the option to make enums available in javascript, similar to:
```C++
enum_<Notation>("Notation")
```
Generally, I'd like to avoid introducing to many enums as there are not default types in JavaScript.
Perhaps, it's best to keep this as a simple int.
In this case, we should give a detailed comment in `Fairy-Stockfish/tests/js/test.js` what this int values mean.

QueensGambit (changes requested):
Thank you very much for your contribution @thearst3rd !
I appreciate the added functionality to **ffish.js**.
Please also have a look at my comments.

Additionally, every function in **ffish.js** requires an inidividual unit test as well as a comment what the function does.
* https://github.com/ianfab/Fairy-Stockfish/blob/master/tests/js/test.js

Having an automated generated documentation page for every function would be preferable of course, but this may be added later.

Moreover, please increment the version counter for **ffish.sj** to **0.6.3**.
* https://github.com/ianfab/Fairy-Stockfish/blob/4f3885107e990adb97f37dcc44b0998bcf5972d5/tests/js/package.json#L3

In general, the **ffish.js** library aims to have a syntax similar to python-chess because python-chess has a large adoption among users and having a similar syntax accelerates the learning curve when trying out **ffish.js**. 
This is not possible for all functions of course because Python is a bit different compared to JavaScript and _Fairy-Stockfish_ supports much more variants, including variants with larger board sizes.

If you like you can add a link to your project to the main readme file of **ffish.js** under a new section called **Example projects**
* https://github.com/ianfab/Fairy-Stockfish/blob/master/tests/js/README.md

src/ffishjs.cpp
QueensGambit: In python-chess there is method called `outcome(*, claim_draw: bool = False) ` and an `enum` called `class chess.Termination(value)`.
It could make sense to use these as well here.

thearst3rd: These are great comments - thank you!

For the `Value`, I was simply copying the way pyffish did it. But I like the way python-chess does it a lot better. I have a few thoughts:

If we wanted an enum, we could create a function `board.outcome()` with the values: `CHECKMATE`, `STALEMATE`, `INSUFFICIENT_MATERIAL`, `N_MOVE_RULE`, `N_FOLD_RULE`, and `VARIANT_END_CONDITION` or something along those lines.

In python-chess, the `chess.Outcome` object has a function `result()` which is a string, so I was also wondering if we wanted to have a function `board.result()` which returns a string in "1-0", "0-1", and "1/2-1/2". I feel like creating a whole `Outcome` object for ffish.js is probably overkill. Though maybe not, I'd love to hear your thoughts!

thearst3rd: On this topic - currently the function `board.isGameOver()` only checks for a lack of legal moves. Should this function be adjusted to also account for conditions such as insufficient material/optional draws?

thearst3rd:
I tweaked the insufficient material functions and added test cases for them. I would like to hear your opinions on how I should proceed with the other functions before digging too deep and making tests.

My thoughts are - rather than implement `board.isImmediateGameEnd` and `isOptionalGameEnd`, we should instead implement `board.result(claimDraw = false)` which returns a string of who won ("1-0", "0-1", "1/2-1/2"). And if we want, we can also implement `board.outcome(claimDraw = false)` which returns an enum of what end condition was reached.

Let me know what you want me to do, and I can do it!

- [ ] Tweak and write tests for insufficient material functions
- [ ] Write a `board.result()` function and test cases
- [ ] Maybe - write a `board.outcome()` function and test cases
- [ ] Maybe - tweak the `board.isGameOver()` to reflect insufficient material/optional

QueensGambit:
Thank you for the quick update @thearst3rd . I like this approach.
As some small adjustments, I suggest to let `board.result(claimDraw = false)` the string `"*"` if the game is still undecided yet. In pgn notation the same notation is used, therefore it should be self-explanatory what it means.

Next, for `board.outcome(claimDraw = false)`, I suggest to add the value `UNDECIDED` to the the enum `Termination`.
I considered to return a string instead of an enum here, but I think returning an enum instead should lead to less error prone code for the user.

The enum values `CHECKMATE`, `STALEMATE`, `INSUFFICIENT_MATERIAL`, `NFOLD_REPETITION` and `VARIANT_END` look good to me. However, I suggest to rename `N_MOVES` into `N_MOVES_LIMIT`.

Regarding, `board.isGameOver()` I recommend to also add the optional bool `claimDraw = false` here.
Testing for insufficient material here, sounds good to me and shouldn't break the code of users who are using `board.isGameOver()`.

I agree that adding `board.isImmediateGameEnd()` and `isOptionalGameEnd()` is no longer required if `board.outcome()` and `board.result()` are available.

thearst3rd:
I've implemented a few of the above comments - I still need to write test cases but I'm heading off to an event so I'll get back to it later tonight.

Some considerations I made: for the enum values, I changed `N_MOVES` to `N_MOVE_RULE` and `NFOLD_REPETITION` to `N_FOLD_REPETITION` since I thought those looked a bit nicer. I was also considering changing `UNDECIDED` for something like `ONGOING`. Not to [bikeshed](https://en.wiktionary.org/wiki/bikeshedding) however - you have the final say in all of these :)

src/ffishjs.cpp
```diff
@@ -263,10 +263,59 @@ class Board {
...
+
+  std::string result(bool claim_draw) const {
+    Value result;
+    bool gameEnd = false;
+    if (is_insufficient_material()) {
+      gameEnd = true;
+      if (pos.material_counting())
+        result = pos.material_counting_result();
+      else
+        result = VALUE_DRAW;
```
thearst3rd, line 278: I think this whole function should be reviewed to make sure I am complying with all of the rules that affect variants, since I don't know most of them :) But this part in particular - does this make sense? I'm not exactly sure when material counting comes into play. I figure that most of the weird end-game conditions are handled by `pos.is_immediate_game_end()` but I want to make sure that this is correct.

I will probably need more help when I write the `board.outcome()` function, because there, I actually need to know why the game ended.

@ianfab

QueensGambit:
Hi @thearst3rd ,
I approve the following changes:
* `N_MOVES` -> `N_MOVE_RULE`
* `NFOLD_REPETITION` -> `N_FOLD_REPETITION`
* `UNDECIDED` -> `ONGOING`

src/ffishjs.cpp
ianfab: I would like to avoid duplicating variant logic as much as possible, since it is very complex, and it should mostly be defined in `Position` methods as the core game logic. Therefore I think the draw result adjudication should be moved to a `Position` method, which anyway would simplify the code, since this very same logic regarding material counting already is duplicated in three places (my bad^^). It could e.g. be something like `Value Position::draw_result(int ply = 0)`.

The outcome method is somewhat problematic, since implementing it would either duplicate logic, or complicate the position adjudication if integrated there. The game end methods could of course return an enum instead of a bool, which likely should not have any impact on performance or functionality, although I am still not sure if this added complexity is acceptable considering that the core engine does not need this information at all. So I am a bit undecided on this topic at the moment. Also @gbtami might be interested in this discussion since pychess-variants currently has custom logic outside of pyffish to determine the type of game end for displaying purposes.

gbtami: Yea, pychess has its own code (with lots of variant specific parts) to determine game results. See https://github.com/gbtami/pychess-variants/blob/master/server/game.py#L516

thearst3rd: These are great comments, thanks guys.

Regarding draw adjudication, I figured out that my current method doesn't even work correctly (if the variant is set to armageddon and you get a stalemate, the result is `1/2-1/2` instead of `0-1`). That might be easily fixable in this case, just set the result to `VALUE_DRAW` on insufficient material, then before returning, check if the current value is `VALUE_DRAW` and if so, factor in material counting. Of course, as you said, it may be nicer to have a function somewhere in `Position`, I'd be happy to take a crack at it though I'm not sure I'll do it right :)

I checked out the pychess-variants code, and it's definitely more complicated than I thought it would be :smile: I'm also realizing it's not fully complete, for instance it just checks if the variant name is "atomic" when seeing if a pseudo-royally checkmated piece should count as checkmate rather than stalemate. How can we know that generically, for all similar variants? Of course, that's fine for pychess.org since that's the only variant that handles that case. But if we want it in a generic library, there is extra complexity.

Also, there are probably some refactors we would want to make for optional draws. AFAIK, n-fold repetition and n-move rule (and others) are all implemented in one lumped-together method `Position::is_optional_game_end()`. How can we check which of those end conditions was reached unless we re-implement them ourselves? My guess is it would be best to split them into different functions (maybe `Position::is_optional_game_end_n_fold()`, `Position::is_optional_game_end_n_moves()`, etc), then check for all of those functions in `Position::is_optional_game_end()`. That way, that function still works as intended, and we can now check for the individual conditions when we need them (like in ffish.js). That's an idea.

thearst3rd:
Based on the comments above, I feel like implementing a nice `board.outcome()` method is going to be very hard in the current state. I'm pretty confident we can get the `board.result()` into an accurate state however, since the available functions implemented in `Position` implement most of the hard details.

thearst3rd:
> If you like you can add a link to your project to the main readme file of **ffish.js** under a new section called **Example projects**

Done :smile: Let me know if the formatting/content is ok!

src/ffishjs.cpp
QueensGambit: > The outcome method is somewhat problematic, since implementing it would either duplicate logic, or complicate the position adjudication if integrated there. The game end methods could of course return an enum instead of a bool, which likely should not have any impact on performance or functionality, although I am still not sure if this added complexity is acceptable considering that the core engine does not need this information at all. So I am a bit undecided on this topic at the moment. Also @gbtami might be interested in this discussion since pychess-variants currently has custom logic outside of pyffish to determine the type of game end for displaying purposes.

I share your concern about duplicate code functionality.
Do you see a way of moving some of these methods to e.g. [apiutil.h](https://github.com/ianfab/Fairy-Stockfish/blob/9205b0839984c5c9f5afa5ce6022de68b9a3e5a7/src/apiutil.h) or to let them depend on already existing methods in a more principled way? I think these kinds of new methods would greatly improve the user experience of the library.

QueensGambit:
> > If you like you can add a link to your project to the main readme file of **ffish.js** under a new section called **Example projects**
> 
> Done 😄 Let me know if the formatting/content is ok!

Thank you for adding your demo project to the readme.
Having a small interactive demo hello world app for **ffish.js** is awesome.

I would have worded the last sentence a little differently, but that is also more a matter of preference.

> A simple ffish.js + chessgroundx website made by thearst3rd, mostly just to mess around with both libraries.

-> A simple ffish.js + chessgroundx website made by thearst3rd that features the core functionality of both libraries.

src/ffishjs.cpp
thearst3rd: I had the idea to add `Termination &outcome` to methods like `Position::is_immediate_game_end`, right next to the `Value &result`. That way, in the function where it's being asked, we can optionally output the `Termination` at the same time as the `Value`. So, something like:

```c++
bool Position::is_immediate_game_end(Value& result, int ply) const {
    Termination outcome;
    return Position::is_immediate_game_end(result, outcome, ply);
}

bool Position::is_immediate_game_end(Value& result, Termination& outcome, int ply) const {
    // ...
    outcome = VARIANT_END; // or CHECKMATE, etc
    // ...
}
```

Apply a similar thing to all functions that output a `Value` representing the end of the game.

Does that sound feasible/reasonable?

QueensGambit:
@thearst3rd , @Fulmene : I have just published your changes in a new **ffish.js** release on npm. :tada: 
* https://www.npmjs.com/package/ffish

PR #369, opened 2021-09-24
Remove Sf README.md

SriMethan:


ianfab:
Thanks for your suggestion. Is there a particular reason why you suggest to remove that? The reason why I kept it so far is because a lot of that information also applies to Fairy-Stockfish, so if I removed it I should add back the relevant parts in some other form, and that seemed like quite some work to me. I understand it might be slightly confusing to have it there unchanged from official stockfish, but tried to mitigate that by clearly separating the sections. If you have a suggestion how to better integrate the relevant information from the official Stockfish readme in the Fairy-Stockfish documentation (e.g., readme or wiki), that is very welcome.

SriMethan:
Ok

PR #374, opened 2021-09-30
Pawnsideways

mtaktikos:
A branch for the Kramnik chess variant Pawnsideways

### DIFF
--- a/src/misc.cpp
+++ b/src/misc.cpp
@@ -160,7 +160,7 @@ string engine_info(bool to_uci, bool to_xboard) {
 
   if (!to_xboard)
     ss << (to_uci  ? "\nid author ": " by ")
-       << "Fabian Fichter, Michael Taktikos";
+       << "Fabian Fichter, pawnsideways by Michael Taktikos";
 
   return ss.str();
 }
--- a/src/misc.cpp
+++ b/src/misc.cpp
@@ -160,7 +160,7 @@ string engine_info(bool to_uci, bool to_xboard) {
 
   if (!to_xboard)
     ss << (to_uci  ? "\nid author ": " by ")
-       << "Fabian Fichter";
+       << "Fabian Fichter, Michael Taktikos";
 
   return ss.str();
 }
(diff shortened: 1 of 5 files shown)


ianfab:
In order for search to work properly, you also need to adjust `Position::pseudo_legal`, which validates moves from the transposition table. Otherwise you will likely see strange effects in cases where a sideways pawn move is the best move, since it gets into a loop of finding the best move and rejecting it again. Another option is to generate the sideways pawn moves as `<SPECIAL>` moves, because then a simpler but slower validation is done, which works out of the box. Fixing pseudo legal move validation should significantly improve playing strength, in my experience typically >100 Elo.

mtaktikos:
On 2021-10-01 17:05, Fabian Fichter wrote:
>
> In order for search to work properly, you also need to adjust 
> |Position::pseudo_legal|, which validates moves from the transposition 
> table. Otherwise you will likely see strange effects in cases where a 
> sideways pawn move is the best move, since it gets into a loop of 
> finding the best move and rejecting it again. Another option is to 
> generate the sideways pawn moves as |<SPECIAL>| moves, because then a 
> simpler but slower validation is done, which works out of the box. 
> Fixing pseudo legal move validation should significantly improve 
> playing strength, in my experience typically >100 Elo.
>

Thanks for the suggestion, the idea with "special" is probably reachable 
for my unperfect programming skills, will give them a try.
However, even so as it is, the playing strength doesn't seem to be a 
problem. The sparring partner Nebiyu is typically an engine which 
reaches in chess about 2400 - 2500 Elo,
and here against Fairy-SF it can not survive even the opening, therefore 
I guess Fairy-SF is > 1000 Elo stronger

> The files are shown as added instead of updated. You need to update 
> them under /src, not at the root.

Oh. Will try it again now. I have no experience with github and hope to 
master this task.

ianfab:
I was not aware that CI tests are even run for drafts. Please use `[skip ci]` in your commit messages in order to avoid failing tests, or alternatively close the PR since referring to your branches in principle is also sufficient.

PR #376, opened 2021-10-01
Branch for the Kramnik chess variant Pawnback

mtaktikos:
Because of alternative pawns, do not chose standard chess, it will play also the pawnback variant


ianfab:
The files are shown as added instead of updated. You need to update them under /src, not at the root.

PR #381, opened 2021-10-03
Expose position.fen's showPromoted and countStarted

Fulmene:


ianfab:
Looks good to me. @QueensGambit Any objections?

src/ffishjs.cpp
```diff
@@ -175,6 +175,11 @@ class Board {
     return this->pos.fen();
   }
 
+  std::string fen(bool showPromoted, int countStarted) const {
+    countStarted = std::min<unsigned int>(countStarted, INT_MAX); // pseudo-unsigned
```
QueensGambit, line 179: I checked the source of _Fairy-SF_ and I think you can safely delete this line.

The method `counting_ply()` already applies all required checks:
https://github.com/ianfab/Fairy-Stockfish/blob/2f49208dd18124e1126454a8fbcf6ea997cd3495/src/position.h#L1178-L1180

Moreover, applying the min-operator by comparing with `INT_MAX` is not necessary, unless I'm missing something.

tests/js/test.js
```diff
@@ -200,6 +200,13 @@ describe('board.fen()', function () {
   });
 });
 
+describe('board.fen(showPromoted, countStarted)', function () {
+  it("it returns the current position in fen format, modified according to additional parameters", () => {
```
QueensGambit, line 204: For me the sentence _modified according to additional parameters_ seems not to sufficiently explain what these additional parameters are supposed to do.
How are promoted pieces indicated using `showPromoted`? After checking the source code, I guess using a `~` symbol after the corresponding piece.
I assume `countStarted` sets the no-progress counter manually, in the FEN.

Can you make the comment more explicit?  e.g.
```js
"it returns the current position in fen format. Promoted pieces are followed by the symbol '~' if `showPromoted` is true. The parameter `countStarted` overwrites the non-progress counter."
```

tests/js/test.js
```diff
@@ -200,6 +200,13 @@ describe('board.fen()', function () {
...
+    let board = new ffish.Board("makruk", "8/6ks/3M~2r1/2K1M3/8/3R4/8/8 w - 128 18 50");
+    chai.expect(board.fen(true, -1)).to.equal("8/6ks/3M~2r1/2K1M3/8/3R4/8/8 w - 128 0 50");
```
QueensGambit, line 206: If we give just one example, it might be confusing and people might think that setting `countStarted` to negative values is what users are supposed to do.
I suggest to add at least a single different example with a positive `countStarted` value.

QueensGambit (changes requested):
Hello @Fulmene ,
thanks for you contribution! Adding this functionality to **ffish.js** makes sense.
I only have some small change requests.

QueensGambit:
I also just checked how **python-chess** does it.
In **python-chess** there is also the `en_passant` parameter available:
```
 fen(*, shredder: bool = False, en_passant: Literal[legal, fen, xfen] = 'legal', promoted: Optional[bool] = None) → str[source]
 ```
 https://python-chess.readthedocs.io/en/latest/core.html#chess.Board.fen
 
 However, I think adding `en_passant` is not mandatory because I think it does comply to _Fairy-SF_ very well.

QueensGambit (approved):
Thank you for the changes. It looks good to me now. :+1: 

The only thing that is missing before publishing a new **ffish.js** version to npm is to increment the version counter. I think it makes sense to publish a new release after pull-request #362 has been successfully merged.

PR #393, opened 2021-11-05
Simplify configuration of custom royal pieces

### LINKED ISSUE #363: Confusion about the "kingType" parameter when reading from variants.ini
I hope you don't mind my onslaught of PRs/issues :)

I'm a bit confused about the how "kingType" works when reading from variants.ini. This is partly because there are no examples of its use in the given variants.ini.

Lets say I want to make a variant that is the same as chess, except the king moves as a centaur. To do this in C++ is quite easy, it looks like this:

```c++
    Variant* centaurking_variant() {
        Variant* v = chess_variant_base()->init();
        v->kingType = CENTAUR;
        return v;
    }
```

I am trying to figure out the equivalent way of doing such a variant in ini format. I'm not sure what to set the kingType as - it parses it by reading in the character from the pieceToChar string - but there isn't a piece character for centaur. I've been trying stuff like:

```
[centaurking:chess]
centaur = c
kingType = c
```

~~But that hasn't worked (the king still moves like a king)~~_(edit: it works, I had a typo before, now fixed)_. Additionally, it's defining a "centaur" piece mapped to the letter c that I'm not planning on using. It seems like "kingType" does two different things in c++ and ini - I'm not sure if this is a bug/missing feature, or if I am just misunderstanding something.

Let me know if I did something wrong, or if there is actually something that needs to be changed in Fairy-Stockfish.

ianfab:
Use the same syntax for custom king as for other custom pieces.

Closes #363.


ianfab:
@thearst3rd What do you think about this solution?

thearst3rd:
Looks good to me!

A few drawbacks that I'm sure you've considered: this breaks compatibility with any preexisting variant which using the kingType parameter. Shouldn't be a big issue as long as you highlight that on the releases page or whatever.

Also, it reduces the number of custom pieces by one. Not that I'd expect many people to need that many :)

Assuming all that is fine, I think this is good! I'll try it out soon. Thanks!!

src/parser.cpp
```diff
@@ -415,7 +429,8 @@ Variant* VariantParser<DoCheck>::parse(Variant* v) {
                 std::cerr << "Can not use kings with blastOnCapture." << std::endl;
             if (v->flipEnclosedPieces)
                 std::cerr << "Can not use kings with flipEnclosedPieces." << std::endl;
-            // We can not fully check custom king movements at this point
+            // We can not fully check support for custom king movements at this point,
+            // since custom pieces are only initialized on loading of the variant.
```
thearst3rd, line 433: Maybe add something like "We will assume this is valid, but it might cause problems later if it's not". That would clear it up for me

src/variants.ini
```diff
@@ -86,7 +86,7 @@
 
 ### Custom pieces
 # Custom pieces can be defined by using one of the available slots:
-# customPiece1, customPiece2, ..., customPiece26
+# customPiece1, customPiece2, ..., customPiece25
 # E.g., pawns without double steps could be described as:
 # customPiece1 = p:mfWcfF
```
thearst3rd, line 91: Somewhere around here we should document:

```
# You can define a custom king in the same way you can define another custom piece:
# E.g., to make a king that moves like a centaur:
# king = k:KN
```

You also created the `centaurking` example which is good :+1: But the top of this document is the closest I could find to a "language specification" for the variants.ini, so additionally putting a note here would be useful for documenting that.

ianfab:
Thanks for your feedback. Yes, those two drawbacks are hard to avoid, but I think should be fine. Your comments should be addressed now.

thearst3rd:
That was fast! Looks all good to me :)

Thanks!

PR #394, opened 2021-11-05
Remove redundancy from S-Chess FEN

### LINKED ISSUE #83: Redundancy in S-Chess FEN
S-Chess FENs with castling rights have redundant gating information, e.g., the starting position is `rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR[EHeh] w KQABCDEFGHkqabcdefgh - 0 1` instead of `rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR[EHeh] w KQBCDFGkqbcdfg - 0 1`.

**This does not have any impact on game play**, but only results in non-standard FEN output of the `d` command for S-Chess.

ianfab:
Closes #83.


ianfab:
@gbtami This would change the starting position FEN for S-Chess, but remaining compatible in both directions. Still, it might have some impact on pychess to change the FEN string, so please let me know if you have any concerns.

PR #415, opened 2021-11-30
Fix search for option aliases in UCI dialects

### LINKED ISSUE #412: Support hashsize in UCCI
I can't set size of hash table in Chinese Xiangqi GUIs I tested. UCCI uses option `hashsize(spin)` instead of `Hash(spin)`.

cloudfish:
Commit b01c9ab does not fix #412, so I tried to debug it. This should fix #412 now.

### DIFF
--- a/src/uci.cpp
+++ b/src/uci.cpp
@@ -115,7 +115,7 @@ namespace {
 
     if (Options.count(name))
         Options[name] = value;
-    // UCI dialects do not allow spaces
+    // Deal with option name aliases in UCI dialects
     else if (is_valid_option(Options, name))
         Options[name] = value;
     else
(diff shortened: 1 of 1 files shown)


ianfab:
Thanks. When I briefly tested it yesterday it seemed to work, but I will re-check. I wanted to avoid using the equality operator in order to consider case insensitivity, but forgot that actually the equality operator anyway already is overriden to consider that, so I guess your solution should be fine.

src/uci.cpp
```diff
@@ -568,7 +568,7 @@ bool UCI::is_valid_option(UCI::OptionsMap& options, std::string& name) {
   std::string protocol = options["Protocol"];
   for (const auto& it : options)
   {
-      if (options.key_comp()(option_name(it.first, protocol), name))
+      if (option_name(it.first, protocol) == name)
```
ianfab, line 571: Since we are comparing strings here and not an Option to a string, we actually have to use the case insensitive less comparator instead of the equality operator in order to avoid having a case sensitive match. I think my version was wrong because the comparator is testing for "less", not for equality. So something like the below should work I think.
```suggestion
      std::string optionName = option_name(it.first, protocol);
      if (!options.key_comp()(optionName, name) && !options.key_comp()(name, optionName))
```

PR #422, opened 2021-12-22
Support checking for checks on pseudo-royal pieces

Fulmene:
This adds checking for checks on pseudo-royal pieces like in Atomic to pyffish's gives_check and ffish.js's isCheck functions.

### DIFF
--- a/src/apiutil.h
+++ b/src/apiutil.h
@@ -358,6 +358,11 @@ inline bool has_insufficient_material(Color c, const Position& pos) {
     return true;
 }
 
+inline bool is_check(const Position& pos) {
+    return pos.checkers()
+        || (pos.extinction_pseudo_royal() && pos.attackers_to_pseudo_royals(~pos.side_to_move()));
+}
+
 namespace FEN {
 
 enum FenValidation : int {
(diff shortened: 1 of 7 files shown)


Fulmene:
It just occurred to me, but would it be better to just check whether the variant has pseudo-royal mechanic within pos itself and do away with the flag in the API?

Something like
```
// ffishjs.cpp
bool isCheck() {
    bool pseudoRoyal = pos.extinction_pseudo_royal(); // or some other check
    if (pseudoRoyal)
        ...
}
```

I'm not sure if the pseudo-royal mechanic is mutually exclusive to regular royal pieces but this seems like a better approach.

ianfab:
Thanks. Yes, checking for `extinction_pseudo_royal` instead of having a flag sounds like a good idea, at least unless there is a specific scenario where checking only for checks on kings is strictly necessary. Royal and pseudo-royal are not mutually exclusive (see e.g. coregal), so I think it would be good to always check for `checkers`, since it anyway is fast, and then additionally checking for pseudo-royal attacks only if the rules say so.

Since the logic in pyffish and ffishjs is identical, I would suggest to move it to `apiutil.h` to reuse it. We could or maybe should then also consider to use this in the SAN generation (regarding `+`) in order to make PGNs more compatible, e.g., for atomic. However, this IMO does not need to be decided now, it is just a possibility opened up by this change.

PR #428, opened 2022-01-10
Create a workflow for releases

TheYoBots:
The build.yml file is renamed to release.yml and it Uploads both linux and windows binaries that can be published in Releases.


ianfab:
Thanks!

PR #434, opened 2022-01-30
Enable Release workflow on all branches

SriMethan:


ianfab:
Thanks, but I do not think this is really needed.

PR #440, opened 2022-02-13
Atomic check exception

Fulmene:
I made it so a pseudo-royal piece is not considered to be attacked if capturing it would blast ALL of the attacker's pseudo-royal pieces. For regular Atomic, this should be enough. This certainly is not covering many edge cases for multiple pseudo-royals though.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -923,8 +923,13 @@ Bitboard Position::attackers_to(Square s, Bitboard occupied) const {
 Bitboard Position::attackers_to_pseudo_royals(Color c) const {
   Bitboard attackers = 0;
   Bitboard pseudoRoyals = st->pseudoRoyals & pieces(~c);
+  Bitboard opponentPseudoRoyals = st->pseudoRoyals & pieces(c);
   while (pseudoRoyals) {
       Square sr = pop_lsb(pseudoRoyals);
+      if (blast_on_capture()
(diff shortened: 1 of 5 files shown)


src/position.cpp
```diff
@@ -923,8 +923,13 @@ Bitboard Position::attackers_to(Square s, Bitboard occupied) const {
 Bitboard Position::attackers_to_pseudo_royals(Color c) const {
   Bitboard attackers = 0;
   Bitboard pseudoRoyals = st->pseudoRoyals & pieces(~c);
+  Bitboard pseudoRoyalsTheirs = st->pseudoRoyals & pieces(c);
   while (pseudoRoyals) {
       Square sr = pop_lsb(pseudoRoyals);
+      if (blast_on_capture()
+          && (pseudoRoyalsTheirs & attacks_bb<KING>(sr)) == pseudoRoyalsTheirs)
```
ianfab, line 930: By the definition of st->pseudoRoyals it is already considered that losing **one** of them would lose the game, see https://github.com/ianfab/Fairy-Stockfish/blob/67af9946b06c73eace9bbe7dcf771204cd7cb304/src/position.cpp#L543-L553
Therefore, it should be sufficient here to check that `(pseudoRoyalsTheirs & attacks_bb<KING>(sr))` is truthy.

Fulmene: I see. So pseudo-royalness only applies for the last piece before extinction kicks in anyway.

ianfab:
Thanks

PR #441, opened 2022-02-13
Make castling through flying general illegal

### LINKED ISSUE #429: Fairy allowing castling into king-faceoff.
https://www.pychess.org/YdJt8zts
https://www.pychess.org/v8LdTGzB

Also, Idk if it's a bug or the intended behaviour, but it allows to castle when the king passes through a square where it would be facing off enemy king.

Fulmene:
Fixes #429


ianfab:
Thanks, merged. The CI failure is a false positive, will likely be resolved by my latest commit.

PR #443, opened 2022-02-14
Simplify releases workflow

TheYoBots:


.github/workflows/release.yml
```diff
@@ -8,60 +8,39 @@ on:
 
 jobs:
   windows:
-    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        arch: ["x86-64-bmi2", "x86-64-modern", "x86-64"]
+    runs-on: windows-latest
```
ianfab, line 14: I prefer to use stable identifiers (like windows-2019) to avoid scenarios where CI can suddenly fail without code changes, e.g. when github updates what is `latest`.

TheYoBots: I've updated the workflow.

TheYoBots:
Sorry I hadn't replied earlier. I've changed the indentifier to `windows-2022` and it builds binaries for both Windows and Linux.

PR #448, opened 2022-03-21
Adding Breakthrough for smaller boards

TheYoBots:
Based on https://github.com/ianfab/Fairy-Stockfish/discussions/242

Adds Breakthrough 5x5,6x6,7x7

Sorry if I'm supposed to get the bench value, I wasn't sure which variants to set it to, to get the bench value.

### DIFF
--- a/src/variant.cpp
+++ b/src/variant.cpp
@@ -848,6 +848,36 @@ namespace {
         v->blackFlag = Rank1BB;
         return v;
     }
+    Variant* breakthrough5_variant() {
+        Variant* v = breakthrough_variant();
+        v->maxRank = RANK_5;
+        v->maxFile = FILE_E;
+        v->startFen = "ppppp/ppppp/5/PPPPP/PPPPP w 0 1";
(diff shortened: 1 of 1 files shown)


ianfab:
Thanks, but I doubt these are worth supporting as built-ins, since they are not practically relevant, but only might be of scientific interest with respect to solving attempts. And for this purpose support via config file should be sufficient, therefore closing it.

Regarding your question about bench, the one that is referred in commit messages is the output of `./stockfish bench` of the non-largeboard version. However, for commits that do not change standard chess functionality you do not need to repeat it, since it should be unchanged.

TheYoBots:
Sod do I submit the same Pull request, just with the variant configured in the variants.ini file? because I found the wiki of solved games extremely confusing considering that it speaks about breakthrough 5x5 being solved, but does mention what changes need to be made or how it can be configured in the variants.ini file.

ianfab:
Yes, adding them to the variants.ini is an option.

PR #460, opened 2022-04-05
Bump pyffish version

gbtami:


ianfab:
If this update is about the latest fix, that one does not affect pyffish/ffishjs, since it only can be triggered in search (or qsearch to be specific).

PR #465, opened 2022-04-19
Update Variants.ini

Belzedar94:
Add support for many new variants.

### DIFF
--- a/src/variants.ini
+++ b/src/variants.ini
@@ -1,5 +1,5 @@
 # Fairy-Stockfish, a UCI chess variant playing engine derived from Stockfish
-# Copyright (C) 2018-2022 Fabian Fichter
+# Copyright (C) 2018-2021 Fabian Fichter
 #
 # Fairy-Stockfish is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -70,7 +70,7 @@
 # gold (WfF)
 # dragonHorse (BW)
 # clobber (cW)
-# breakthrough (fmWfF)
+# breakthrough (fWfFcF)
 # immobile ()
 # cannon (mRcpR)
 # janggiCannon (pR)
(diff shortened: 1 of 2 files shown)


ianfab:
Thanks. To me those anti/coffee- and -house variants are a bit much though, since the default variants.ini should mainly contain relevant and diverse variants, so that a variety of configuration scenarios are covered with a limited set of variants as examples. So I would prefer to not add those hybrid variants that are very redundant, and mainly include the ones that are unique, like massacre, mounted, etc., especially the ones which are on pychess and/or for which there are NNUE.

Also the variants.ini this PR is based on seems to be a bit outdated, since it would revert a few changes that were done, especially in the documentation at the top and regarding "centaurking". Please update/revert those changes (e.g., lines 89, 171).

src/variants.ini
```diff
@@ -733,6 +927,399 @@ nMoveRule = 0
...
+capturesToHand = false
+whiteDropRegion = *1 *2 *3 *4
+blackDropRegion = *5 *6 *7 *8
+immobilityIllegal = true
+flagPiece = k
+whiteFlag = *8
+blackFlag = *1
+castling = false
+
+[twokings:chess]
```
ianfab, line 1006: Twokings is a variant already defined by several GUIs and engines, and the rules here differ from that, so please either rename and comment it, or remove it, in order to avoid confusion.

src/variants.ini
```diff
@@ -733,6 +927,399 @@ nMoveRule = 0
...
+customPiece2 = b:cB
+customPiece3 = n:cN
+customPiece4 = r:cR
+startFen = rBbnbqNn/bQrnQRRr/rrqbNqQb/bBNqbNBN/bNRrrrBQ/QRQqRnNB/qBRnnqNQ/BBQnqnRR w - - 0 1
+stalemateValue = loss
+extinctionValue = loss
+extinctionPieceTypes = *
+
 # 5x5 breakthrough
 [breakthrough5:breakthrough]
```
ianfab, line 1324: This variant is duplicated now.

Belzedar94:
Totally agree with your comments. I removed most of the hybrids except the ones I found uniques or interesting, and I fix the rest of the things you commented.

src/variants.ini
```diff
@@ -335,12 +302,7 @@ promotionPieceTypes = q
 doubleStep = false
 castling = false
 
-# Centaurking
-# A variant demonstrating how to define a custom royal piece movement
-[centaurking:chess]
```
ianfab, line 340: Please add this one back, because it was added explicitly to show the custom king feature.

src/variants.ini
```diff
@@ -733,26 +801,171 @@ nMoveRule = 0
...
+castling = false
+startFen = rnnbqkbnnr/pppppppppp/10/10/10/10/PPPPPPPPPP/RNNBQKBNNR[] w KQkq - 0 1
+maxRank = 8
+maxFile = 10
+chess960 = false
+promotionRank = 6
+doubleStep = true
+castling = false
+
+[atomicshogi]
```
ianfab, line 918: This looks wrong, at least speaking from the name of the variant. Perhaps the inheritance `:shogi` is just missing. However, even with that it would not work yet, since the usual replacement of king by commoner would still be required to properly turn it into a working atomic variant. I think you at least need to add
```
king = -
commoner = k
extinctionValue = loss
extinctionPieceTypes = k
blastOnCapture = true
extinctionPseudoRoyal = true
```

PR #474, opened 2022-05-12
Fix PGN parsing 960 variants

gbtami:


ianfab:
Thanks. I am not really familiar with how other applications use the "Variant" tag in this regard. E.g., do lichess, cutechess, etc. also encode the "960" in this tag, or do they use something else?

src/ffishjs.cpp
```diff
@@ -562,10 +562,14 @@ Game read_game_pgn(std::string pgn) {
         headersParsed = true;
         auto it = game.header.find("Variant");
         if (it != game.header.end()) {
-          game.variant = it->second;
+          game.is960 = it->second.find("960") != std::string::npos;
+          if (game.is960) {
+            game.variant = it->second.substr(0, it->second.length() - 3);
```
ianfab, line 567: Either the `find` should be restricted to start searching at `size() - 3` or it should erase starting from the `find` position in order to ensure that 960s occuring elsewhere in the string won't lead to weird artifacts like `960chess` -> `960ch`.

gbtami:
Lichess uses [Variant "Chess960"] and can import 960 lichess variant games from PGN if they use Variant tag without "960" for other variants.
Other software usually can't export/import other 960 variants form PGN. Maybe I'm wrong.
This issue raised in https://github.com/gbtami/pychess-variants/issues/693 when people tried to import pychess games exported by pychess itself :)

gbtami:
@QueensGambit may I ask you to push out a new release to npm, please?

QueensGambit:
Done #475.

PR #494, opened 2022-06-03
add brain mode

movepicker:
Brain Mode is A Feature That Make Stronger Chess Engine , The Idea of Brain Mode Is Base Of ChessBase Authors , They Found A Lot Of Process May Cause Reduce CPU Performance & Cause Calculate Wrong Best Move 

_In engine tournaments, if two engines are playing against each other, using permanent brain slows down the side that is computing the next move (because it has to share resources with the other side's permanent brain)_

http://help.chessbase.com/Fritz/15/Eng/index.html?000028.htm

This method is currently used by ChessBase GUI (Fritz , Komodo , ..) As **Permanent Brain**

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -22,6 +22,8 @@
 #include <cstring>   // For std::memset
 #include <iostream>
 #include <sstream>
+#include <chrono>
+#include <thread>
 
 #include "evaluate.h"
 #include "misc.h"
@@ -230,8 +232,10 @@ void MainThread::search() {
   // GUI sends a "stop" or "ponderhit" command. We therefore simply wait here
   // until the GUI sends one of those commands.
 
-  while (!Threads.stop && (ponder || Limits.infinite))
-  {} // Busy wait for a stop or a ponder reset
+  while (!Threads.stop && (ponder || Limits.infinite || brain))
+  {	  
+  if (brain) { std::this_thread::sleep_for(std::chrono::milliseconds(100));  break; } 
+  } // Busy wait for a stop or a ponder reset
 
   // Stop the threads if not already stopped (also raise the stop if
   // "ponderhit" just reset Threads.ponder).
@@ -1944,6 +1948,8 @@ void MainThread::check_time() {
   // We should not stop pondering until told so by the GUI
   if (ponder)
       return;
+  if (brain)
+	  return;
 
   if (   rootPos.two_boards()
       && Time.elapsed() < Limits.time[rootPos.side_to_move()] - 1000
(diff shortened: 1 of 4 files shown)


ianfab:
Pondering already is supported. Also this is a chess variant engine, so compatibility to chessbase products is not relevant here, as they are for standard chess only.

movepicker:
this method  use only for fix depth ,  ponder mode is good but it must side by fix depth  & need Chess GUI Support To Stop It For Each Position

Loop 
`go ponder depth 10`
`stop`

ianfab:
Either way this fork is about chess variants and not about introducing new Stockfish features. The basic functionality we just inherit from official Stockfish by merging from upstream, so if you want to change such a feature you need to get it into official Stockfish (but I can tell you this one won't make it). Please only file PRs specific to the purpose of this project here.

movepicker:
I Explain This Issues 2 Times in Official Stockfish but Its Authors Block (Spam) Me Because They Say This Method is False , You Can See My Issues Post
 
 https://github.com/official-stockfish/Stockfish/issues/4001#issue-1222935210
 
 https://github.com/official-stockfish/Stockfish/issues/4012#issue-1227809931
 
 unfortunately ,  i can't make any contributions in Official Stockfish Project

ianfab:
Please don't misuse this repository as a second chance for rejected ideas. If official stockfish considers something as spam I very likely will too, and if you refuse to acknowledge that I might take similar measures.

movepicker:
please compare this two behavior

the pull request at official-stockfish
 -------------------
 _I have blocked @movepicker for 7 days, after having mentioned multiple times that his interactions with the repo are not appropriate. This is a last warning._

_Originally posted by @vondele in https://github.com/official-stockfish/Stockfish/issues/4039#issuecomment-1140515397_

 the pull request at fairy-stockfish
 -------------------

 
 _Thanks for your suggestion. However, slowing down perft testing doesn't seem like a desirable goal to me. The main purpose of `perft` (at least for me) is to validate and debug move generation, so slowing it down doesn't help there, and having the speedup isn't a bug. I see that for profiling reasons actually traversing the whole tree might be useful in some cases, but in the rare case that this scenario occurs one can always apply your change manually._

_Originally posted by @ianfab in https://github.com/ianfab/Fairy-Stockfish/issues/489#issuecomment-1142515709_

ianfab:
Yes, I try to be polite and I am glad you appreciate that, but this doesn't change the fact that I rejected all your PRs so far because they weren't relevant, and I am not interested in a meta discussion now. If you have something that is **actually relevant and aligns with the goal and purpose of this project**, you can create a PR or write comments. Anything else is just wasting my time, including your last comment. So consider this as a warning.

Issue #416, opened 2021-12-01
Fairy doesn't check if promoted pieces actually exist in a variant when validating FENs.


kezsulap:
Example FENs considered valid
Shinobi: rnb+qkbnr/pppppppp/8/8/8/8/PPPPPPPP/LH1CK1HL[LHMMDJ] w kq - 0 1
Chennis: 1f+km3/1p1s3/7/7/7/3S1P1/3M1F1[] w - - 1 1
Kyotoshogi: p+nks+l/5/5/5/+LS+K+NP[-] w 0 1

kezsulap:
Happens also in variants with no piece promotion at all
Standard chess: rn+bqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1
is also considered valid

ianfab:
For reference, https://github.com/ianfab/Fairy-Stockfish/blob/9022a70549bf741db2fe4b57af42739b1cb91a2d/src/apiutil.h#L470-L480 is the relevant code.

PR #497, opened 2022-06-14
Issue 416

johndoknjas:
This pull request partly solves issue #416. The code now only allows certain special characters, depending on the variant being played. E.g., the "+" char won't be allowed in the seirawan variant. I've also called the FEN::validate_fen function from position.cpp - not sure if this would make sense or if it should be left as is. One thing I was a bit unsure of is whether '~' chars can appear in shogi FENs (and also whether they can appear in makpong FENs). Right now I'm assuming they're only allowed in shogi ones.

Some of the examples in issue 416 still aren't resolved. E.g., Kyotoshogi: p+nks+l/5/5/5/+LS+K+NP[-] w 0 1 is still treated as valid, since it contains no special chars which aren't allowed. It's just that some chars are in the wrong spots in the string.

### DIFF
--- a/src/apiutil.h
+++ b/src/apiutil.h
@@ -845,16 +845,12 @@ inline Validation check_digit_field(const std::string& field) {
 inline std::string get_valid_special_chars(const Variant* v) {
     std::string validSpecialCharactersFirstField = "/";
     // Whether or not '-', '+', '~', '[', ']' are valid depends on the variant being played.
-    if (v->variantTemplate == "shogi")
-    {
+    if (v->shogiStylePromotions)
         validSpecialCharactersFirstField += '+';
-        if (v->pieceDrops)
-            validSpecialCharactersFirstField += '-';
-    }
-    if (v->capturesToHand && v->nnueAlias != "crazyhouse") // Only 2 particular variants of crazyhouse have this alias.
+    if (!v->promotionPieceTypes.empty())
         validSpecialCharactersFirstField += '~';
     if (!v->freeDrops && (v->pieceDrops || v->seirawanGating || v->arrowGating))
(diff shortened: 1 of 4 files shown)


ianfab (changes requested):
Thanks for your PR. Overall it looks quite good, there are just two points I have in my CR comments. One is regarding the FEN validation in the position setup, the other is about the association of the allowed characters with the rules. I assume that the issues with failing tests will already be fixed when my comments are addressed, since the failures look related to them.

johndoknjas:
@ianfab Thanks for the feedback, it's very helpful. I'll work on implementing these changes.

ianfab:
There is one remaining test failure https://github.com/ianfab/Fairy-Stockfish/runs/6897885928?check_suite_focus=true I didn't expect. I think that is because the `conclude()` is only called on insert of variants, but the start FEN validation for custom variants already happens on parsing. So I think we need to call it also before https://github.com/ianfab/Fairy-Stockfish/blob/8ec72eff3bcf82648396756d03085d2c9363257d/src/parser.cpp#L399 or more generally at the beginning of the consistency checks there, as these might use derived properties in general. That is more of a bug in the current master, but didn't show before since we did not use derived properties there yet.

ianfab:
https://github.com/ianfab/Fairy-Stockfish/blob/8ec72eff3bcf82648396756d03085d2c9363257d/src/variant.cpp#L1560-L1569 is the related code. Custom variants are added, but the validation currently already happens directly after (as part of) parsing. It could be considered in the future to change the this, but for now the duplicate call of conclude to me is the simplest solution.

johndoknjas:
@ianfab Ah ok, I see.

PR #503, opened 2022-06-28
Add in-built support for Raazuvaa (Maldivian Chess)

Belzedar94:
I hope I didn't miss any important change to be made

### DIFF
--- a/src/ucioption.cpp
+++ b/src/ucioption.cpp
@@ -48,7 +48,7 @@ namespace UCI {
 std::set<string> standard_variants = {
     "normal", "nocastle", "fischerandom", "knightmate", "3check", "makruk", "shatranj",
     "asean", "seirawan", "crazyhouse", "bughouse", "suicide", "giveaway", "losers", "atomic",
-    "capablanca", "gothic", "janus", "caparandom", "grand", "shogi", "xiangqi"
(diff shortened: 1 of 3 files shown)


src/ucioption.cpp
```diff
@@ -48,7 +48,7 @@ namespace UCI {
 std::set<string> standard_variants = {
     "normal", "nocastle", "fischerandom", "knightmate", "3check", "makruk", "shatranj",
     "asean", "seirawan", "crazyhouse", "bughouse", "suicide", "giveaway", "losers", "atomic",
-    "capablanca", "gothic", "janus", "caparandom", "grand", "shogi", "xiangqi"
+    "capablanca", "gothic", "janus", "caparandom", "grand", "shogi", "xiangqi", "raazuvaa"
```
ianfab, line 51: Standard variants are built-in winboard variants

src/variant.cpp
```diff
@@ -79,6 +79,14 @@ namespace {
         v->add_piece(FERS, 'f');
         return v;
     }
+      // Raazuva (Maldivian Chess)
+    Variant* raazuvaa_variant() {
```
ianfab, line 83: Here you only define the variant. You also need to `add` it below.

ianfab (changes requested):

Belzedar94:
Both issues corrected!

ianfab (approved):

PR #505, opened 2022-07-12
Add in-built support for Miguel Illescas' Dragon Chess.

romstad:
I'm trying to help Miguel Illescas popularise this variant, and adding support to Fairy-Stockfish would be a good step in that direction. I'm planning to train an NNUE evaluation for this variant soon, too.

Phenomenal project, by the way! Back when I started writing the program that eventually became Stockfish, I couldn't in my wildest dream imagine that it would some day develop into something as awesome as this. Thank you so much for your work!

### DIFF
--- a/src/variant.cpp
+++ b/src/variant.cpp
@@ -560,12 +560,14 @@ namespace {
     // square on the back rank instead of making a normal move.
     Variant* dragon_variant() {
         Variant *v = chess_variant_base()->init();
-        v->variantTemplate = "dragon";
+        v->variantTemplate = "bughouse";
         v->pieceToCharTable = "PNBRQ............D...Kpnbrq............d...k";
         v->add_piece(ARCHBISHOP, 'd');
         v->startFen = "rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR[Dd] w KQkq - 0 1";
         v->pieceDrops = true;
         v->capturesToHand = false;
+        v->whiteDropRegion = Rank1BB;
+        v->blackDropRegion = Rank8BB;
         return v;
     }
     // Base used for most shogi variants
--- a/src/variant.cpp
+++ b/src/variant.cpp
@@ -86,7 +86,7 @@ namespace {
         v->castling = false;
         v->doubleStep = false;
         return v;
-    } 
+    }
     // Makruk (Thai Chess)
     // https://en.wikipedia.org/wiki/Makruk
     Variant* makruk_variant() {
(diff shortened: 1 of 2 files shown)


src/variant.cpp
```diff
@@ -552,6 +552,22 @@ namespace {
         v->capturesToHand = true;
         return v;
     }
+    // Dragon Chess
+    // 8x8 variant invented by Miguel Illescas:
+    // https://www.edami.com/dragonchess/
+    // Like regular chess, but with an extra piece, the dragon, which moves like
+    // an archbishop (i.e. bishop+knight). The dragon can be dropped at an empty
+    // square on the back rank instead of making a normal move.
```
ianfab, line 560: The description mentions the piece can only be dropped on the back rank, but according to the rule config it can be dropped anywhere. The relevant parameters for drop regions are:
https://github.com/ianfab/Fairy-Stockfish/blob/67209f150d44b6ac5fcc01659442fcc9a4815e95/src/variant.h#L89-L90

src/variant.cpp
```diff
@@ -552,6 +552,22 @@ namespace {
...
+    Variant* dragon_variant() {
+        Variant *v = chess_variant_base()->init();
+        v->variantTemplate = "dragon";
```
ianfab, line 563: The variantTemplate is only used in communication with Winboard/Xboard (or more generally in the CECP protocol) and refers to built-in variants of the GUI to inherit logic from: https://github.com/ianfab/Fairy-Stockfish/blob/67209f150d44b6ac5fcc01659442fcc9a4815e95/src/variants.ini#L137
Therefore this should equal to one of the predefined template variants in XBoard/Winboard. For most chess based variants `fairy` usually is the recommended choice, but for variants with drops where captured pieces don't return to the opponent's hand (like in crazyhouse/shogi), usually `bughouse` should be used.

romstad:
I have no idea why the Windows build failed here, but since the error was an out-of-memory error during compilation, I guess it can't have anything to do with the rather trivial code changes.

ianfab (commented):
Thanks for your PR. I have added two small review comments.

And thanks for the kind words. Glad to hear that, especially from one of the founders of the project. I am trying my best to leverage the Stockfish code infrastructure for a large space of games, since Stockfish is a way too good project to only be used for one single game :).

ianfab:
Yes, the job failure like was just some random environment related failure unrelated to the PR. I just restarted it, and assume it will succeed.

src/variant.cpp
romstad: If you had asked me, I would have told you that the Stockfish code was too chess-specific and that you would have had better chance of success writing code from the ground up for general chess-like games rather than trying to modify Stockfish.

Fortunately, you didn't ask me. I'm amazed by how well Fairy-Stockfish works even for games very different from chess.

You are right, I messed up the drop rules. I'll fix.

romstad:
I tried fixed the issues you pointed out. I hope it will all work this time!

ianfab:
Thanks, looks good to me. Will likely merge it later today after CI has passed.

PR #522, opened 2022-10-01
Epose variant capturesToHand to js/python

gbtami:


gbtami:
@QueensGambit may I ask you to release a new ffish.js to npm, please?

PR #534, opened 2022-10-17
Add riscv64 support

clefrks:
Adding support for riscv64 architecture. 
Compiled and tested on VisionFive 2 board


ianfab:
Thanks. However, I generally try to avoid maintaining own architecture support and just use the upstream code from official SF. Has this already been suggested and/or merged to official SF?

ianfab:
Thanks for clarifying. This is the "official" SF fork for chess variants (also used on lichess), but not the original base project for standard chess. For that please raise PRs at https://github.com/official-stockfish/Stockfish. I merge upstream changes from there.

PR #537, opened 2022-10-24
enable bit manipulation instruction set 1

niklasf:
A trivial improvement from upstream:

```
bmi1 enables the use of _blsr_u64 for pop_lsb, and is availabe when avx2 is.

verified a small speedup (0.2 - 0.6%)

closes https://github.com/official-stockfish/Stockfish/pull/4202

No functional change
```


ianfab:
Thanks, and sorry for the delay.

PR #566, opened 2023-01-02
Duck

mtaktikos:
Rules update: The stalemated side wins in Duck Chess  into branch ianfab:duck


mtaktikos:
Sorry, this pull request is not intended for the master branch, but for the duck branch.
It is only a single line updating the rules of Duck Chess

PR #567, opened 2023-01-02
Updated rules of Duck Chess

mtaktikos:
The stalemated side wins in Duck Chess,  https://www.chess.com/terms/duck-chess#rules


ianfab:
Thanks for the hint, I added the change to the PR on the "wall" branch. The duck branch is basically obsolete now, I will delete it once the PR is merged.

PR #571, opened 2023-01-13
Add typescript type definition file for ffish.js

Fulmene:


ianfab:
Thanks. Looks fine to me, the failing tests likely were just due to temporary server unavailability. Since my experience with typescript is rather limited I would like to ask @QueensGambit to have a look as well.

gbtami:
I think the type of Board result() should be string (not boolean).

Fulmene:
By the way, I've only tested this file with pychess, which is using the ES6 module version. I'm not sure if this same file would work for the CommonJS version though.

QueensGambit:
@Fulmene Thanks a lot for this contribution. :tada: 
I think having a typescript definition is a valuable addition to the js-library.

We also need an additional unit-test for your fen() function overload, [here](https://github.com/ianfab/Fairy-Stockfish/blob/master/tests/js/test.js#L202).
```js
describe('board.fen(showPromoted)', function () {
```

I think it would be a good idea to also have an automated library documentation for the **ffish.js** library based on your [ffish.d.ts](https://github.com/ianfab/Fairy-Stockfish/pull/571/files#diff-554443b4c4d6e1cfbb2a072a2ec0feda048aa557acd55ba3f70e765df89d688a) file.

Some approaches on how to do this can be found here:
* https://blog.cloudflare.com/generating-documentation-for-typescript-projects/

At the moment, the documentation for the **ffish.js** can be found in the [test.js](https://github.com/ianfab/Fairy-Stockfish/blob/master/tests/js/test.js) file, which is not very user friendly.

QueensGambit:
Thanks!

gbtami:
Add another version bump as well, please!
Thx :)

QueensGambit:
@Fulmene I'm fine with both, but would prefer this to be done within this PR, so I don't need to create two separate versions shortly after each other.

Fulmene:
@QueensGambit Done. Please let me know if the code needs anything else.

QueensGambit:
The code looks good to me. :+1: I'm not an expert in typescript, but I assume I can publish _**ffish.js**_ just as usual given your addition `"types": "ffish.d.ts",` to the **package.json** file.

Fulmene:
All done. 👍🏻

ianfab:
From my side looks ready to merge.

QueensGambit:
**ffish@0.7.3** and **ffish-es6@0.7.3** are now available at https://www.npmjs.com/package/ffish.

Fulmene:
Thank you!

By the way, thoughts on unifying ffish and ffish-es6 into a single package?

QueensGambit:
Yes, that could be a good idea.
Initially we only published **ffish.js**, but had some trouble at making it work in vue.js.
Therefore, we added **ffish-es6.js** as an alternative.
Nowadays, **ffish-es6.js** has about four three times more downloads compared to *ffish.js**.

PR #575, opened 2023-01-15
 Bump ffish version to 0.7.2

QueensGambit:
New package versions are available at:
https://www.npmjs.com/package/ffish


ianfab:
Thanks

PR #579, opened 2023-01-24
static and past gating: isolation, joust, snailtrail

RainRat:
gating could be divided into static gating, and past gating

a couple example variants of how they could be used.

### DIFF
--- a/src/movegen.cpp
+++ b/src/movegen.cpp
@@ -43,6 +43,11 @@ namespace {
             b ^= to - pawn_push(us);
         if (pos.variant()->arrowGating)
             b &= moves_bb(us, type_of(pos.piece_on(from)), to, pos.pieces() ^ from);
+        if (pos.variant()->staticGating)
+            b &= pos.variant()->staticGatingRegion;
+        if (pos.variant()->pastGating)
+            b &= square_bb(from);
+
         while (b)
             *moveList++ = make_gating<T>(from, to, pt, pop_lsb(b));
         return moveList;
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -330,6 +330,9 @@ Variant* VariantParser<DoCheck>::parse(Variant* v) {
     parse_attribute("gating", v->gating);
     parse_attribute("arrowGating", v->arrowGating);
     parse_attribute("duckGating", v->duckGating);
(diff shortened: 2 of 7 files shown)


src/movegen.cpp
```diff
@@ -43,6 +43,11 @@ namespace {
             b ^= to - pawn_push(us);
         if (pos.variant()->arrowGating)
             b &= moves_bb(us, type_of(pos.piece_on(from)), to, pos.pieces() ^ from);
+        if (pos.variant()->staticGating)
```
ianfab, line 46: Please note that move generation and pseudo-legal move validation always need to be in sync in order to ensure that moves coming from the TT are actually valid. Therefore statements analogous to https://github.com/ianfab/Fairy-Stockfish/blob/a0b277d7abfe6094bd44da4b113b00a94f0325a1/src/position.cpp#L1218-L1219 should be added to pseudo-legal validation.

src/parser.cpp
```diff
@@ -330,6 +330,9 @@ Variant* VariantParser<DoCheck>::parse(Variant* v) {
     parse_attribute("gating", v->gating);
     parse_attribute("arrowGating", v->arrowGating);
     parse_attribute("duckGating", v->duckGating);
+	parse_attribute("staticGating", v->staticGating);
```
ianfab, line 333: Very minor thing, but would be nice to replace tab indentation by spaces here.

src/variant.cpp
```diff
@@ -418,6 +418,51 @@ namespace {
         v->stalemateValue = VALUE_MATE;
         return v;
     }
+
```
ianfab, line 421: If you can find reliable sources for descriptions of (some of) the variants, preferably wikipedia, chessvariants.com, etc., it would be nice to put the links to them in comments for each variant, similar to how it is done for some other variants. It isn't a requirement though, just nice to have in case it is available.

ianfab (commented):
Thanks for your PR. The changes generally look good to me, there is just one issue regarding pseudo-legal move validation that definitely needs to be addressed before it can be merged.

ianfab:
Thanks

IoannisKarpasitis:
Hi, may I ask which templates support the staticGating option?

PR #580, opened 2023-01-26
Nuclear

RainRat:
Adds two new options, pawnsGetBlast (pawns can be destroyed in blasts like in Atomic), and pawnsCanPetrify (pawns can trigger petrification).


ianfab:
Thanks for your PR. Generally I am a bit more hesitant about this PR than your previous one since the increase of the space of supportable variants seems very limited compared to the added complexity, which always is the tradeoff that needs to be evaluated in this project.

I think there might be ways to improve it, like converting pawnsCanPetrify into an option that represents the piece type that is immune to petrification, which would both remove the hard-coded piece type (PAWN) in the logic while not adding new code branches. I generally like such kinds of extensions that just generalize existing rules by replacing hard-coded values by configurable values as that usually requires very few changes in logic.

However, currently the code seems to also have functional issues. When compiling with `debug=yes` nuclear already crashes on depth 6. In the presence of royal or pseudo-royal pieces legal move validation usually needs some care.

PR #605, opened 2023-03-25
Update descriptions

gnunicode:
Update description of which lame leapers are supported and update README.md to include new variants


ianfab:
Thanks!

PR #611, opened 2023-03-27
move isolation, joust, snailtrail to default build. fix typos

RainRat:
I put isolation, joust, snailtrail under ALLVARS at first to be on the safe side. I haven't seen any problems, so am requesting enabled by default. The most intensive would be Isolation 7x7, which has 360 possible moves at most, which still looks ok.


ianfab:
Thanks, looks good to me.

Just as a general comment I usually don't fix typos in comments originating from official SF to reduce merge conflicts when merging from upstream, but I obviously won't ask you to put the typos back in now that you fixed them :), and of course the ones that are Fairy-SF specific should be fixed anyways. I will merge it once the tests have passed.

PR #614, opened 2023-03-29
 Bump ffish version to 0.7.4

QueensGambit:
New package versions are available at:
https://www.npmjs.com/package/ffish


gbtami:
@ianfab Possible there is no need to keep appveyor CI now after we have wheels CI.

PR #622, opened 2023-04-01
Add Board.attackedPieces() to ffish.js

Fulmene:


ianfab:
Thanks. Can be merged either way, but just out of curiosity, do you have any specific use case in mind?

Fulmene:
The use case is displaying attacks on each royal piece separately for variants with multiple of them on one side, like Spartan. The idea is to inform chessgroundx which squares are under attack so that only the royal pieces within those squares are highlighted. Do you have ideas for improvements that could make this cover more use cases?

Fulmene:
I actually just found out about pseudoRoyalCandidates yesterday. Am I correct to assume that's the bitboard of the pieces that would've become pseudo-royal when the number drops to the extinction threshold? If so, I think I would like to make use of that instead considering I'm currently unhappy with how I need to loop through all the pieces and still filter them again inside chessgroundx.

ianfab:
In that case I would suggest to change `attackers_to_pseudo_royals` to make it return the attacked instead of the attackers, and then there could be another method in ffishjs very analogous to is_check just returning the squares instead of a boolean, along the lines of e.g. (untested):

```
inline bool is_check(const Position& pos) {
    return pos.checkers()
        || (pos.extinction_pseudo_royal() && pos.attacked_pseudo_royals(pos.side_to_move()));
}

inline Bitboard checked(const Position& pos) {
    return (pos.checkers() ? pos.square<KING>(pos.side_to_move()) : Bitboard(0))
        | (pos.extinction_pseudo_royal() ? pos.attacked_pseudo_royals(pos.side_to_move()) : Bitboard(0));
}
```

Since I already implemented duple check in the `attackers_to_pseudo_royals` yesterday, it should only require minor changes there to change the returned bitboard (attacked instead of attackers), then it should work already for Spartan.

Fulmene:
Thanks. In that case, I'll implement it in a separate PR.

PR #627, opened 2023-04-03
Update Variants.ini

Belzedar94:
Add support for many new variants.

Any hint on how to easily change all the deprecated PromotionRank options to the new PromotionRegion[Color]?

By the way, I took into account all the comments from the previous PR, so for instance, I tried to only include the anti/coffee variants that I saw interesting/popular or playable on an up-to-date website.

Thanks for reviewing!

### DIFF
--- a/src/variants.ini
+++ b/src/variants.ini
@@ -860,7 +860,7 @@ blastOnCapture = true
 castling = false
 extinctionOpponentPieceCount = 1
 
-# Hybrid of atomic and zh. Zh is th base variant. --> SOLVED
+# Hybrid of atomic and Crazyhouse. --> SOLVED
 [atomiczh:atomic]
 dropChecks = false
 pieceDrops = true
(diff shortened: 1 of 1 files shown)


src/variants.ini
```diff
@@ -853,3 +853,428 @@ stalemateValue = loss
...
+extinctionValue = loss
+extinctionPseudoRoyal = true
+extinctionPieceTypes = k
+extinctionPieceCount = 1
+startFen = rnbkkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBKKBNR w - - 0 1
+
+# Anti-Antichess
+# Lose at anti-chess to win at anti-antichess.
+# Variant defined in Liantichess website. Credits to SriMethan for the definition.
+[anti_antichess:giveaway]
```
ianfab, line 1051: I was so far always using dashes as separators in variant names, so for consistency would suggest the same here.

ianfab (commented):
Thanks. Regarding the migration of promotionRank, there is nothing automated, one currently needs to manually convert it, similar as I did in https://github.com/fairy-stockfish/Fairy-Stockfish/commit/b1124aec70c0ea3ba1db3b7e7a2c72b36adae467#diff-33c5d3b051dea86f192d5315580529b171d25cb80f51c5c9db0fb2c603129846.

Some of those liantichess variants still seem a bit random to me (e.g. coffeerace, antiminishogi), but if you had a look at them and find them interesting in their own right, I am fine with adding them.

Belzedar94:
Updated the PR with your comments. Hope it's all good already :)

Regarding Coffeerace and Antiminishogi, I still found it interesting to add them. Coffeerace because even though it's solved (I'll update the wiki soon), it's playable and I found some fun in Liantichess. Similar case with Antiminishogi, I found it funny to play in Liantichess as it's really different than the rest of antis I tried.

ianfab:
Thanks, the warnings regarding promotionRank are now all gone. The CI failure is due to remaining warnings of the INI checker about the pieceToCharTable, see https://github.com/fairy-stockfish/Fairy-Stockfish/actions/runs/4606059358/jobs/8138887520. You should be able to reproduce that by running the `check variants.ini` locally. Maybe it is a bit too strict to require a valid pieceToCharTable for all variants in the default INI, but would still be nice to have if possible.

PR #628, opened 2023-04-03
add kono and fox_and_hounds

RainRat:
Kono is apparently a draw with perfect play ( https://www.whatdowedoallday.com/five-field-kono ).

Fox and Hounds is solvable with Stockfish but has history and multiple variants so could be interesting to have.

### DIFF
--- a/src/variant.cpp
+++ b/src/variant.cpp
@@ -547,22 +547,23 @@ namespace {
         v->maxRank = RANK_5;
         v->maxFile = FILE_E;
         v->reset_pieces();
-        v->add_piece(CUSTOM_PIECES, 'p', "mF"); //diagonally, no capture
+        v->add_piece(CUSTOM_PIECE_1, 'p', "mF"); //diagonally, no capture
         v->startFen = "ppppp/p3p/5/P3P/PPPPP w - - 0 1";
-        v->flagPiece = CUSTOM_PIECES;
+        v->flagPiece = CUSTOM_PIECE_1;
         v->flagRegion[WHITE] = make_bitboard(SQ_A5, SQ_B5, SQ_C5, SQ_D5, SQ_E5, SQ_A4, SQ_E4);
         v->flagRegion[BLACK] = make_bitboard(SQ_A1, SQ_B1, SQ_C1, SQ_D1, SQ_E1, SQ_A2, SQ_E2);
-        v->numFlag = 7;
+        v->flagPieceCount = 7;
         return v;
     }
 
     Variant* fox_and_hounds_variant() { //https://boardgamegeek.com/boardgame/148180/fox-and-hounds
         Variant* v = chess_variant_base()->init();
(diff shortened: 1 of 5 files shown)


src/variant.cpp
```diff
@@ -542,6 +542,31 @@ namespace {
...
+    }
+
+    Variant* fox_and_hounds_variant() { //https://boardgamegeek.com/boardgame/148180/fox-and-hounds
+        Variant* v = chess_variant_base()->init();
+        v->reset_pieces();
+        v->add_piece(CUSTOM_PIECES, 'h', "mfF"); //Hound
+        v->add_piece(CUSTOM_PIECES+1, 'f', "mF"); //Fox
+        v->startFen = "1h1h1h1h/8/8/8/8/8/8/4F3 w - - 0 1";
+        v->stalemateValue = -VALUE_MATE;
+        v->flagRegion[WHITE] = Rank8BB;
```
ianfab, line 566: I think the flagPiece is missing here

src/variant.h
```diff
@@ -141,6 +141,7 @@ struct Variant {
   int extinctionOpponentPieceCount = 0;
   PieceType flagPiece = NO_PIECE_TYPE;
   Bitboard flagRegion[COLOR_NB] = {};
+  int numFlag = 1;
```
ianfab, line 144: Similar attributes are named ...Count or ...PieceCount, so for consistency I would suggest to use a similar naming here (e.g., flagCount or flagPieceCount).

src/variant.cpp
```diff
@@ -1775,6 +1800,8 @@ void VariantMap::init() {
     add("isolation", isolation_variant());
     add("isolation7x7", isolation7x7_variant());
     add("snailtrail", snailtrail_variant());
+    add("fox_and_hounds", fox_and_hounds_variant());
```
ianfab, line 1803: So far underscores aren't used in official variant names in Fairy-SF, so I would suggest to use either a dash or no separator. There isn't any specific reason to avoid underscores, e.g., there isn't a technical limitation, it is just about the existing (unwritten) naming convention.

ianfab (commented):
Thanks for your PR. I wasn't familiar with these games, but it is nice that they can be supported with so little change. I just have a few small change requests, other than that it looks good to me.

src/variant.cpp
RainRat: Yes. I also noticed the naming convention for custom pieces had changed so I changed that too.

src/variant.cpp
RainRat: Done.

RainRat:
This error, I don't think my change caused it. Is it something you're already aware of?

ianfab:
Thanks for addressing the comments, looks ready for merge to me now. The CI error looks like some temporary failure, I just retriggered the job, and once it passes I will merge.

PR #638, opened 2023-04-11
Flagvariants

RainRat:
- Add flagPieceBlockedWin (for flagPieceCount >1). If at least one of your flag piece is in the target flag zone, and all spaces in the flag zone are occupied, then you win.
- Moved kono to variants.ini (to more easily try variations of flagPieceBlockedWin)
- Support flagPieceCount in variants.ini
- Add Xhess, Buffalo, Knight's Halma in variants.ini
- Mention Kono and "Fox and Hounds" in readme.
- Typo fixes


RainRat:
I looked over the Zillions library (breakthrough/race types in specific) and did not find another where blocked piece=win. Usually the rules obligate a move or adjudicate it a different way. If one really wanted to deep dive, they could look into some of the other games that obligate a move and come up with some different generalizations for that.

So, currently, the criteria is if at least one flag piece is in the flag zone and none of the squares were *empty*.
-If you had a game with flag pieces and non-flag pieces, AND you got one flag piece in the zone AND you filled the rest with your non-flag pieces, then this option would adjudicate as a win. This is admittedly not ideal.
-Wall squares are not empty squares, so they would be considered blocking.
 
I can possibly look at more games to find more ways to generalize, but with just one example, It's hard to generalize.

RainRat:
1. I did the suggested optimization.
2. I fixed the merge conflict
3. I did a survey of breakthrough variants to see if there were other variations of the "counting enemy pieces as your own" rule and as far as I could find, the Kono rule is unique and I don't see any other variations on that rule that need support.
[breakthrough survey.txt](https://github.com/fairy-stockfish/Fairy-Stockfish/files/11256333/breakthrough.survey.txt)

If you decide not to merge it, I can put the other proposed changes into a new branch and propose just those.

ianfab:
Thanks a lot for the research and updates. Functionality-wise it looks good to me, but I have some concerns regarding the implementation of the game end conditions. In general it is good practice keep things separate and explicit, so what you wrote makes sense. However, Fairy-Stockfish is very feature heavy and subject to further extension, so I try to reduce the complexity in terms of number of code paths as much as possible, even if that makes the individual code paths maybe a bit harder to read, as it can otherwise quickly get out of hand with many rule configuration options when you try to allow arbitrary combination of rules. This is quite specific to this project and maybe also debatable, but so far this works best for me in this project to keep it maintainable.

Therefore I would like to rather implement the conditions in a similar (though not necessarily identical) way as my formulas/pseudo-code above. However, I appreciate your efforts and don't like to ask you now to rewrite your functional code. So depending on what you prefer you could either try to update it yourself according to what I am describing if you are interested, or I would also be fine to merge it as is, and then would do the refactoring myself at a later point.

RainRat:
It's your project so it's up to you. If I had some concrete examples of rules that needed support, I'd be inclined to revise, but I looked high and low and couldn't find anything similar.

ianfab:
I noticed that I already introduced some inconsistencies with respect to naming and also that `*` (all pieces) is not allowed for flag variants inconsistently with e.g. extinction variants, so it needs a general refactoring, therefore I merged the PR as is.

PR #652, opened 2023-04-24
add All Queens, Rooksquare, Alapo, A Simple Game, Picaria, Nine Holes

RainRat:


ianfab:
Thanks. The changes in general look good, just `connectOrthoOnly` seems a bit too specific to me. Would be nice to make the option more flexible to allow configuring the directions, though currently we do not have any config option type for a set of directions yet. However, unless there is another idea how to make this more flexible without having the verbosity of multiple options (like `connectOrthoOnly` and `connectDiagOnly`), I think adding a configurable set of directions probably should be the way to go.

src/position.h
```diff
@@ -203,7 +203,11 @@ class Position {
   bool flag_reached(Color c) const;
   bool check_counting() const;
   int connect_n() const;
-  bool connect_ortho_only() const;
+  bool connect_horizontal() const;
+  bool connect_vertical() const;
+  bool connect_diagonal() const;
+
+  const std::vector<Direction>& connect_directions() const;
```
ianfab, line 210: This seems to be unused

RainRat: Yep, it was

PR #663, opened 2023-06-21
add Atomar. fix typo of "weight"

RainRat:


ianfab:
Thanks! One limitation of this PR is that the new features can not be combined with pseudo-royal pieces, but I assume that was intentional. I will try to have a closer look at the review in the next days.

src/variants.ini
```diff
@@ -166,6 +166,8 @@
 # mandatoryPiecePromotion: piece promotion (and demotion if enabled) is mandatory [bool] (default: false)
 # pieceDemotion: enable demotion of pieces (e.g., Kyoto shogi) [bool] (default: false)
 # blastOnCapture: captures explode all adjacent non-pawn pieces (e.g., atomic chess) [bool] (default: false)
+# blastImmuneTypes: pieces completely immune to explosions (even at ground zero) [PieceSet] (default: none)
+# diplomacyTypes: pieces that can't capture another piece of same types (e.g., kings (commoners) in atomar) [PieceSet] (default: none)
```
ianfab, line 170: I would suggest a more descriptive name here, e.g., like `mutuallyImmuneTypes`. Not sure if there is a better suggestion, I would just like to make the name less specific to atomar.

src/parser.cpp
```diff
@@ -357,6 +357,8 @@ Variant* VariantParser<DoCheck>::parse(Variant* v) {
     parse_attribute("mandatoryPiecePromotion", v->mandatoryPiecePromotion);
     parse_attribute("pieceDemotion", v->pieceDemotion);
     parse_attribute("blastOnCapture", v->blastOnCapture);
+    parse_attribute("blastImmuneTypes", v->blastImmuneTypes, v->pieceToChar);
```
ianfab, line 360: Since `blastImmuneTypes` is not supported in combination with pseudo-royalty yet, it would be good to add a warning in the consistency checks below to make this explicit.

ianfab (commented):
Thanks, and sorry for the long delay. I just added two small review comments.

PR #693, opened 2023-08-04
Cannonshogi

gbtami:
Fixes https://github.com/fairy-stockfish/Fairy-Stockfish/issues/304


ianfab:
Thanks. Out of curiosity, did you try out the variant? Just asking because I assume it should in theory work fine, but it is a quite extreme combination of cannon pieces not present in any variant so far.

gbtami:
Yes. You can try it using pychess cannonshogi branch https://github.com/gbtami/pychess-variants/tree/cannonshogi

ianfab:
Thanks

PR #695, opened 2023-08-14
add Atomar. Optimize connect.

RainRat:


src/position.h
```diff
@@ -206,6 +208,7 @@ class Position {
   bool connect_horizontal() const;
   bool connect_vertical() const;
   bool connect_diagonal() const;
+  std::vector<Direction> getConnectDirections() const;
```
ianfab, line 211: I would suggest to return a constant reference here (similar as e.g. for the strings above), since it showed in the past that for containers (vectors, sets, ...) this improves performance quite a bit as it avoids copying a non-trivial object. Other than that the changes look good to me.

ianfab:
Thanks

PR #698, opened 2023-08-23
Add variant Dodo Chess

M-DinhHoangViet:


src/variant.cpp
```diff
@@ -344,6 +344,19 @@ namespace {
         v->checking = false;
         return v;
     }
+    // Dodo Chess
+    // https://www.evochess.com/chess-variant/dodo
+    Variant* dodochess_variant() {
+        Variant* v = chess_variant_base()->init();
```
ianfab, line 350: Since dodochess is basically racing kings with a different starting position, I would suggest to inherit from racing kings here. Then only the differences would need to be explicitly mentioned, i.e., the startFen, as well as optionally removing pawns and queens from the set of piece types.

src/variants.ini
```diff
@@ -1018,6 +1018,16 @@ flagRegionBlack = *1
 flagMove = true
 checking = false
 
+# Dodo Chess
+# Win by campmate. No checks allowed.
+# Variant defined in Pychess discord and previously playable with Chessboi bot.
+[dodo:chess]
```
ianfab, line 1024: I do not have a strong opinion on whether to add dodo as a built-in or a configured variant, but I would prefer to have it only as one of the two (whichever you prefer) to avoid redundancy.

ianfab (commented):
Thanks for your PR. I just added two small CR comments.

ianfab:
Sure, ultimately maintainers have the last say on their project, but ideally the code should IMO be improved through interaction. If the maintainer just overwrites whatever code is thrown at him, there hardly is any learning on either side. It of course is up to you whether are you willing to put in the effort to address the comments, and if you don't that is ok as well, but IMO it is more realistic to submit non-trivial PRs with the expectation that there will be comments to be addressed.

PR #703, opened 2023-08-25
add Nuclear

RainRat:
The interaction between Blast and Petrify:

If no special effort, then each piece in the blast will be petrified. This differs from Nuclear rules, where only ground zero is petrified. So petrifyBlastPieces was made to toggle this. It is set to true by default, in case anyone was depending on the previous behaviour.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -2003,7 +2003,7 @@ void Position::do_move(Move m, StateInfo& newSt, bool givesCheck) {
           }
 
           // Make a wall square where the piece was
-          if (var->petrifyOnCapture)
+          if ((var->petrifyOnCapture) && (var->petrifyBlastPieces || bsq==to))
           {
               st->wallSquares |= bsq;
               byTypeBB[ALL_PIECES] |= bsq;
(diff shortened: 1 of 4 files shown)


src/position.cpp
```diff
@@ -2003,7 +2003,7 @@ void Position::do_move(Move m, StateInfo& newSt, bool givesCheck) {
           }
 
           // Make a wall square where the piece was
-          if (var->petrifyOnCapture)
+          if ((var->petrifyOnCapture) && (var->petrifyBlastPieces || bsq==to))
```
ianfab, line 2006: As far as I understand it in this case `petrifyBlastPieces` will only have an effect if `petrifyOnCapture` is true.

However, in principle I think whether the capturing piece and whether the blast pieces are petrified could be independent of each other. There are 4 possible combinations of the two (captured piece petrified yes/no, blasted pieces petrified yes/no), but with this implementation we can only cover 3 of them, as the scenario of an unpetrified capturing piece but petrified blasted pieces is not possible. So I would suggest to make it so that the two do not interfere with each other. I guess something like `if (bsq == to ? var->petrifyOnCapture : var->petrifyBlastPieces)` could work.

I don't think that the scenario I mentioned is particularly relevant, but if we can support it basically without adding any code complexity that would be nice to have.

RainRat: added your suggestion. it works.

it's no extra code to keep the previous behaviour, so i didn't change it. but it's your project so up to you.

ianfab:
Thanks!

PR #705, opened 2023-08-29
add Duck region mask. add ducktest. add forward/checkers

RainRat:
not expecting ducktest to actually be added, just for testing.

### DIFF
--- a/src/variant.h
+++ b/src/variant.h
@@ -110,7 +110,7 @@ struct Variant {
   bool duckGating = false;
   bool staticGating = false;
   bool pastGating = false;
-  Bitboard staticGatingRegion[COLOR_NB] = {AllSquares, AllSquares};
+  Bitboard gatingRegion[COLOR_NB] = {AllSquares, AllSquares};
   bool seirawanGating = false;
   bool cambodianMoves = false;
   Bitboard diagonalLines = 0;
--- a/src/variant.h
+++ b/src/variant.h
@@ -108,10 +108,9 @@ struct Variant {
   bool gating = false;
   bool arrowGating = false;
   bool duckGating = false;
-  Bitboard duckRegion[COLOR_NB] = {AllSquares};
   bool staticGating = false;
   bool pastGating = false;
-  Bitboard staticGatingRegion = AllSquares;
+  Bitboard staticGatingRegion[COLOR_NB] = {AllSquares};
   bool seirawanGating = false;
   bool cambodianMoves = false;
   Bitboard diagonalLines = 0;
--- a/src/variant.h
+++ b/src/variant.h
@@ -110,7 +110,7 @@ struct Variant {
   bool duckGating = false;
   bool staticGating = false;
   bool pastGating = false;
-  Bitboard staticGatingRegion[COLOR_NB] = {AllSquares};
+  Bitboard staticGatingRegion[COLOR_NB] = {AllSquares, AllSquares};
   bool seirawanGating = false;
   bool cambodianMoves = false;
   Bitboard diagonalLines = 0;
--- a/src/variant.h
+++ b/src/variant.h
@@ -106,11 +106,11 @@ struct Variant {
   int dropNoDoubledCount = 1;
   bool immobilityIllegal = false;
   bool gating = false;
-  bool arrowGating = false;
-  bool duckGating = false;
-  bool staticGating = false;
-  bool pastGating = false;
-  Bitboard gatingRegion[COLOR_NB] = {AllSquares, AllSquares};
+  bool arrowWalling = false;
+  bool duckWalling = false;
+  bool staticWalling = false;
+  bool pastWalling = false;
+  Bitboard wallingRegion[COLOR_NB] = {AllSquares, AllSquares};
   bool seirawanGating = false;
   bool cambodianMoves = false;
   Bitboard diagonalLines = 0;
--- a/src/variant.h
+++ b/src/variant.h
@@ -108,6 +108,7 @@ struct Variant {
   bool gating = false;
   bool arrowGating = false;
   bool duckGating = false;
+  Bitboard duckRegion[COLOR_NB] = {AllSquares};
   bool staticGating = false;
   bool pastGating = false;
   Bitboard staticGatingRegion = AllSquares;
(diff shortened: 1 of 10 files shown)


src/variant.h
```diff
@@ -110,7 +110,7 @@ struct Variant {
   bool duckGating = false;
   bool staticGating = false;
   bool pastGating = false;
-  Bitboard staticGatingRegion = AllSquares;
+  Bitboard staticGatingRegion[COLOR_NB] = {AllSquares};
```
ianfab, line 113: I know it is just for testing, but just FYI this way of initialization does not work. If you initialize `= {0}` that will set all members to zero, but not because this is a general "fill", but because all remaining values will be initialized to the default/zero value. So this is equivalent to `{AllSquares, 0}`, and not `{AllSquares, AllSquares}` if that was the intention.

RainRat:
Let me know what you would like me to do with this: Whether you think duckRegion is usable, and also whether you like it as part of staticGatingRegion or whether it should be its own parameter. Also, I can leave ducktest in(possibly under a different name) if it makes sense to have it as a demo.

You're the boss, so let me know which of the changes you want.

RainRat:
To address the comment, made these changes:
- "staticGatingRegion*" to "gatingRegion*" (still affects both duck and isolation types)
- "gatingRegion" sets bitboards for both black and white

Also removed ducktest because I don't feel strongly about it.

My concern is that the "gating" term feels less descriptive since "gating" is used for two different things, Seirawan "Gating", and dropping walls.

Maybe dropping walls could move away from the "gating" term and use a more specific term. "wallingRegion" or "wallDropRegion".

ianfab:
Thanks. Yes, the "gating" terminology is mainly coming from the implementation rather than the concept in general, so renaming it to be better understandable from a user perspective makes sense. So feel free to rename the usages of "gating" to "walling" in all contexts where it is specific to wall squares.

Belzedar94:
Instead of adding ducktest, I propose to directly use Chennisball as a demo, since it was the main trigger for this change.

Definition:

`[chennisball]
maxRank = 7
maxFile = 7
mobilityRegionWhiteCommoner = b1 c1 d1 e1 f1 b2 c2 d2 e2 f2 b3 c3 d3 e3 f3 b4 c4 d4 e4 f4
mobilityRegionBlackCommoner = b4 c4 d4 e4 f4 b5 c5 d5 e5 f5 b6 c6 d6 e6 f6 b7 c7 d7 e7 f7
customPiece1 = p:fmWfceF
customPiece2 = m:K
cannon = c
commoner = k
fers = f
soldier = s
bishop = b
knight = n
rook = r
promotionPieceTypes = -
promotedPieceType = p:r f:c s:b m:n
promotionRegionWhite = *1 *2 *3 *4 *5 *6 *7
promotionRegionBlack = *7 *6 *5 *4 *3 *2 *1
startFen = 1fkm3/1p1s3/7/7/7/3S1P1/3MKF1[] w - 0 1
pieceDrops = true
capturesToHand = true
pieceDemotion = true
mandatoryPiecePromotion = true
dropPromoted = true
castling = false
stalemateValue = loss
duckGating = true
duckRegionWhite = *5 *6 *7
duckRegionBlack = *1 *2 *3
extinctionValue = loss
extinctionPieceTypes = k
extinctionPseudoRoyal = true`

ianfab:
I have strong doubts about chennisball being supported, I don't think you can combine the duck with drops.

RainRat:
Drops seem to work. Note that I had to change your INI to refer to "duckWalling" and "wallingRegion" to work with this branch. I will add your INI to this proposal.
```

position startpos moves f1e2+,e2a1 b6b5+,b5a1 e2e3-,e3a1 d6d5+,d5a1 d2e2+,e2a1 b5b3-,b3a1 e3d2+,d2a1 b3b2+,b2a1 d2d7-,d7a1 c7d7,d7f3
d

 +---+---+---+---+---+---+---+
 |   | f |   | k |   |   |   |7   [f]
 +---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |6
 +---+---+---+---+---+---+---+
 |   |   |   |+s |   |   |   |5
 +---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |4
 +---+---+---+---+---+---+---+
 |   |   |   |   |   | * |   |3
 +---+---+---+---+---+---+---+
 |   |+p |   |   |+S | P |   |2
 +---+---+---+---+---+---+---+
 |   |   |   | M | K |   |   |1 * [M]
 +---+---+---+---+---+---+---+
   a   b   c   d   e   f   g

Fen: 1f1k3/7/3+s3/7/5*1/1+p2+SP1/3MK2[Mf] w - - 0 6
Sfen: 1f1k3/7/3+s3/7/5*1/1+p2+SP1/3MK2 b Mf 11
Key: FA84B42BEC15CA34
Checkers:
Chased:
position startpos moves f1e2+,e2a1 b6b5+,b5a1 e2e3-,e3a1 d6d5+,d5a1 d2e2+,e2a1 b5b3-,b3a1 e3d2+,d2a1 b3b2+,b2a1 d2d7-,d7a1 c7d7,d7f3 +M@c3,c3c4
d

 +---+---+---+---+---+---+---+
 |   | f |   | k |   |   |   |7 * [f]
 +---+---+---+---+---+---+---+
 |   |   |   |   |   |   |   |6
 +---+---+---+---+---+---+---+
 |   |   |   |+s |   |   |   |5
 +---+---+---+---+---+---+---+
 |   |   | * |   |   |   |   |4
 +---+---+---+---+---+---+---+
 |   |   |+M |   |   |   |   |3
 +---+---+---+---+---+---+---+
 |   |+p |   |   |+S | P |   |2
 +---+---+---+---+---+---+---+
 |   |   |   | M | K |   |   |1   []
 +---+---+---+---+---+---+---+
   a   b   c   d   e   f   g

Fen: 1f1k3/7/3+s3/2*4/2+M4/1+p2+SP1/3MK2[f] b - - 0 6
Sfen: 1f1k3/7/3+s3/2*4/2+M4/1+p2+SP1/3MK2 w f 12
Key: FFA3A52B90A168A2
Checkers:
Chased:

```

RainRat:
taking chennisball out. just focusing on what does work.

ianfab:
Thanks. Gating and drops occupy the same parts of the move encoding, that is why I believe they can't work together. Perhaps we should better document such limitations in the INI validation, it is just hard for me to write all of them down proactively since they usually only cross my mind when I think of the respective combination of rules, which mostly happens only during implementation and when someone proposes a specific variant.

PR #710, opened 2023-09-07
Opposite castling, maybe.

RainRat:


ianfab:
Thanks. I somehow thought this might not easily be possible, but since you usually can't gain castling rights during a game immediately removing the castling right avoids the necessity to keep track of the castling side, that's nice.

PR #717, opened 2023-09-09
add Castle Chess (yes, that's the name of the variant)

RainRat:


ianfab:
Thanks. The condition seems extremely specific to me though, so a generalization would be nice. Since there already are the bitwise castling rights masks I think one could just have such a mask as the configuration value, parse it from something like `castlingWins = KQkq` and check the win condition as `mask & castlingRightUsedByMove`. The parsing will need a bit of logic, but after that the logic should perhaps be no more difficult than if just supporting the specific condition.

RainRat:
This is for the situation where either castling is the win condition
castlingWins = kq

If one castling is prevented, say by capturing the rook should it be a loss? I think the expected answer is no, if the other side castling is still available. So changed the logic.

ianfab:
Was wondering about that too, but your solution looks fine to me. I will just try to slightly simplify it with some syntactic sugar, and make the castling rights a dedicated config type to improve the validation.

PR #719, opened 2023-09-12
flagPieceSafe, Squatter

RainRat:
add flagPieceSafe option. rework Alapo to use it.

add Squatter as another use of flagPieceSafe.

### DIFF
--- a/src/position.h
+++ b/src/position.h
@@ -953,7 +953,7 @@ inline bool Position::flag_reached(Color c) const {
   bool simpleResult = 
         (flag_region(c) & pieces(c, flag_piece(c)))
         && (   popcount(flag_region(c) & pieces(c, flag_piece(c))) >= var->flagPieceCount
-        || (var->flagPieceBlockedWin && !(flag_region(c) & ~pieces())));
+            || (var->flagPieceBlockedWin && !(flag_region(c) & ~pieces())));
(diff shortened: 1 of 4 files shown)


src/position.h
```diff
@@ -950,9 +950,39 @@ inline bool Position::flag_move() const {
...
-  return   (flag_region(c) & pieces(c, flag_piece(c)))
+  bool simpleResult = 
+        (flag_region(c) & pieces(c, flag_piece(c)))
         && (   popcount(flag_region(c) & pieces(c, flag_piece(c))) >= var->flagPieceCount
             || (var->flagPieceBlockedWin && !(flag_region(c) & ~pieces())));
+      
+  if (simpleResult&&var->flagPieceSafe)
+  {
+      Bitboard piecesInFlagZone = flag_region(c) & pieces(c, flag_piece(c));
+      int potentialPieces = (popcount(piecesInFlagZone));
```
ianfab, line 961: Since both attackers_to and legality checking are very expensive I would suggest to have an early return if potentialPieces >= var->flagPieceCount + 1, as already reasoned in your comment. With that also the logic becomes simpler, because once you found an attack you can immediately return false, since potentialPieces == var->flagPieceCount - 1 already is implicit by then.

PR #720, opened 2023-09-12
Add Gethenian variant to INI file

autocorr:
Hi, I've added a custom variant of my own design to the default configuration INI file and would be delighted if it could be included with the default distribution of Fairy-Stockfish.

I have done quite a bit of testing so far, so don't expect to need to revise the rules in the near future (i.e., fairy bit of play-testing, starting position balanced with NN and MCTS).

Clear skies,
Brian


ianfab:
Thanks. If you have already done play-testing that is fairy good.

PR #721, opened 2023-09-14
add Antimatter, Snort

RainRat:


ianfab:
Thanks. Antimatter however doesn't work, since similar to atomar and the like the explosion of pieces needs to be reflected in legality checking and check detection in order for it to be compatible with royal or pseudo-royal pieces. A simple example of that is the position `rnbqkb1r/ppp2ppp/5n2/3p4/4P3/8/PPP1QPPP/RNB1KBNR b KQkq - 0 4`, where dxe4 should not be legal. Therefore I think it would be good to split the PR so that the less problematic variant can already be merged.

ianfab:
Unfortunately the code changes add a bit too much complexity to me in comparison to the added value. Especially considering that as I mentioned gives_check would need to be updated as well. I can see the possibility that after a refactoring of blast and petrification logic this might be possible to introduce with less complexity, but in the current state I am a bit too afraid of maintainability to merge antimatter.

RainRat:
Yeah fair. Removed it, it was more complicated than I thought it would be.

PR #722, opened 2023-09-17
add Edge Walling rule, suggesting refactor of Walling Rule,

RainRat:


ianfab:
Thanks, that's really nice to have this simplification in.

PR #727, opened 2023-09-24
Update README.md

M-DinhHoangViet:


ianfab:
Thanks, but I don't want to link to inofficial bots, especially if there is no related information in the profile page. We could add a link to playstrategy itself though, as they anyway use Fairy-SF for play with the computer.

PR #728, opened 2023-09-25
add 'wall or move' rule (for Atlantis)

RainRat:


src/movegen.cpp
```diff
@@ -443,6 +444,12 @@ namespace {
         // Workaround for passing: Execute a non-move with any piece
         if (pos.pass() && !pos.count<KING>(Us) && pos.pieces(Us))
             *moveList++ = make<SPECIAL>(lsb(pos.pieces(Us)), lsb(pos.pieces(Us)));
+
+        //if "wall or move", generate walling action with null move
+        if (pos.variant()->wallOrMove)
+        {
+            moveList = make_move_and_gating<NORMAL>(pos, moveList, Us, SQ_A1, SQ_A1);
```
ianfab, line 451: I think this might potentially be problematic. Currently passes rely on that a piece makes a non-move, but if no piece is placed on a1 this could potentially break here, or at least the `do_move` code has not explicitly been made ready yet for a non-piece non-move.

ianfab:
Let me know if there is anything I can help with regarding this PR. I have no problem with keeping this open if you plan to return to it at a later point, just want to briefly clarify what is the status.

RainRat:
In case you haven't seen the new commit, I believe I have fixed the problem.

ianfab:
Sorry for the long delay. When I run a search on the starting position for atlantis with debug=yes then it raises a failed assertion at depth 1:
```
stockfish: position.cpp:1032: bool Stockfish::Position::legal(Stockfish::Move) const: Assertion `is_ok(m)' failed.
```

Could you please have a look?

RainRat:
I wasn't able to reproduce this error.

* I merged the latest changes in an odd way, but it should still be understandable.

ianfab:
Here are the exact commands I used to reproduce the error:
```
make clean
make -j2 ARCH=x86-64-bmi2 build debug=yes
./stockfish load variants.ini
setoption name UCI_Variant value atlantis
position startpos
go
```

RainRat:
ok, i think this might be the fix.

ianfab:
Thanks

PR #731, opened 2023-09-29
Add Cfour-anyside, Symphony, Teeko

RainRat:


ianfab (commented):
Thanks, I just added two CR comments.

I only realized thanks to your PR that actually dropOnTop might be a very good candidate to merge into the enclosingDrop. Since the former was implemented first I think I missed the similarity when introducing  enclosing drops. Not necessarily a request to implement this in your PR, just a side note.

RainRat:
> Thanks, I just added two CR comments.
> 
> I only realized thanks to your PR that actually dropOnTop might be a very good candidate to merge into the enclosingDrop. Since the former was implemented first I think I missed the similarity when introducing enclosing drops. Not necessarily a request to implement this in your PR, just a side note.

I was actually thinking of suggesting that next.

RainRat:
1. Implemented optimization suggestions
2. Added code for how "enclosingDrop = top" could work.
3. Manually resolved conflict with 3 previous variants.ini additions.

ianfab:
Thanks!

PR #742, opened 2023-10-24
Update wheels.yml

gbtami:
Python3.7 already reached EOL.


ianfab:
Thanks. The error still seems to be there though. Documentation says
> CPython 3.12 is built by default using Python RCs, starting with cibuildwheel 2.15.

so maybe we need to upgrade cibuildwheels?

gbtami:
I was hoping it will fix it. Let see updating cibuildwheel can solve it.

ianfab:
Thanks, looks like it worked.

PR #746, opened 2023-11-21
Cfour-misere, per-color passing

RainRat:
1. cfour-misere (All connection win conditions can use connectValue. I suppose it's possible for a variant to have one type of connection a loss and another type a win. but that is a stretch and can be handled as it comes up.)
2. Per-color passing for Marco's variant (no variant included to use it specifically)


ianfab:
Thanks, looks good

PR #755, opened 2024-01-30
add Three Musketeers. needs collinearN and connectPieceTypes.

RainRat:


src/variant.h
```diff
@@ -149,12 +149,14 @@ struct Variant {
   bool flagPieceSafe = false;
   bool checkCounting = false;
   int connectN = 0;
+  PieceSet connectPieceTypes = CHESS_PIECES;
```
ianfab, line 152: I think we should either set the default to NO_PIECE_SET, so that no useless calculation is done for non-connect variants, or set it to all piece types and in that case adjust the logic to narrow it down to actually used piece types.

The latter option is what I tend to prefer in recent times, because instead of forcing the user to do consistent configuration, it tries to handle conflicting information gracefully. E.g., one could set the default to all piece types, but then in `Variant::conclude` & the connection piece types with the actually used piece types, and set it to no piece types if none of the connection options are used. This way all of the existing connection variants would still work without changing the INI and at the same time non-connection games would be handled cleanly.

ianfab (commented):
Thanks. I left a few CR comments. I like the addition of the connectPieceTypes, seems like a fairly natural generalization.

src/position.cpp
```diff
@@ -2845,6 +2854,36 @@ bool Position::is_immediate_game_end(Value& result, int ply) const {
...
+          Bitboard b = allPieces;
+          while (b) {
+              Square s = pop_lsb(b);
+
+              int total_count = 1; // Start with the current piece
+
+              // Check in both directions
+              for (int sign = -1; sign <= 1; sign += 2) {
+                  Bitboard shifted = shift(sign * d, square_bb(s));
+                  while (shifted) {
```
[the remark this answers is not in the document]
RainRat: I gave some thought to how it would work with connectDirections, especially the diagonal ones on a non-square board, and couldn't come up with something reasonable.

src/variant.h
RainRat: I did the second, start with all pieces, then optimize to either no pieces, or pieces actually in play.

src/variant.cpp
```diff
@@ -2064,6 +2064,17 @@ Variant* Variant::conclude() {
         connect_directions.push_back(SOUTH_EAST);
     }
 
+    // If not a connect variant, set connectPieceTypes to no pieces.
+    if ( !(connectRegion1[WHITE] || connectRegion1[BLACK] || connectN || connectNxN || collinearN) )
+    {
+          connectPieceTypes = NO_PIECE_SET;
+    }
+    //Otherwise optimize to pieces actually in the game.
+    else if (connectPieceTypes == CHESS_PIECES)
```
ianfab, line 2073: I think it is more generic to simply have
```
else
    connectPieceTypes = connectPieceTypes & pieceTypes;
```
This way you just ignore what is not used and keep what is used.

ianfab:
Thanks for the updates, looks good to me now. Will merge it after CI has passed.

Issue #19, opened 2019-03-11
Hostage Chess?


GregNeto:
Thanks for the program, love it to play around with the exotic variants!

Have you considered to implement Hostage Chess?
https://en.wikipedia.org/wiki/Hostage_Chess
I prefer this to crazyhouse, can be played over the board.

ianfab:
Thanks for your feedback.

The hostage exchange would be tricky to implement, so I doubt it will be feasible in the near term. Furthermore, the rule for checks by pawns that are about to promote is quite ugly (for implementation, and in my opinion it is also unintuitive), which further complicates Hostage chess, but I will keep this issue open as a feature request.

If anyone else is interested in implementing it, feel free to create a pull request.

GregNeto:
You are absolutely right, the promotion rule is the biggest obstacle explaining the rules.
That castling is possible with a dropped rook is also strange.
But the exchange- and droprule is great, me and friends have created our own chess variant from this rule which we enjoy playing very much (also with a different promotion rule)

ianfab:
Although I have never played Hostage chess so far, the game dynamic sounds interesting, but currently Xiangqi is highest on my priority list, which will require a lot of effort.

Since I pondered a bit about the potential issues of a future implementation, as a note for myself I will list what came to my mind:
- gives_check, legal, pseudo_legal, and move generation need to take care of not treating "stuck" pawns on the 7th rank as threats to the king, and some captures have to be recognized as illegal due to providing exchange hostages.
- Check `MAX_MOVES` and stack size issues due to large number of moves (at most 64 * (5 + 2 * 4 + 2 + 1) = 1024 piece drops + ~200 moves -> 1024 moves might be sufficient, i.e., ~1MB stack)
- The second piece type of the hostage exchange could be encoded in the same bits as the "pt_in_hand" for normal piece drops?
- Handle normal piece drops and hostage exchanges as separate move types? How to avoid code duplication?

cab101:
Hello.
I made some implementation of hostage chess, but i can't push my branch. 
I attach 3 patch. 
[patches.zip](https://github.com/fairy-stockfish/Fairy-Stockfish/files/14228439/patches.zip)

PR #758, opened 2024-02-10
Hostage #19

cab101:
Add extra hand - 'prison'.
Captured pieces go to it.
Player can make exchanges by rules in variant (map array: type -> piece set).
Support block pawn promotion and check.
Disabled by default.
Doesn't effect other variants if disabled.


ianfab:
Thanks. As this is a rather big PR it will likely take some time until I will be able to give feedback, but I will try have a look.

PraseodymiumSpike:
very nice to see this variant

PR #759, opened 2024-02-12
Add test case for cannonshogi flying silver cannon

gbtami:


ianfab:
Thanks. Did you intentionally skip the scenario where it currently would fail, namely allowing diagonal captures beyond 2 squares, to make it mergeable before the fix? I am asking because I could either first merge this and then add the test case that the fix solves in the fix itself, or first merge the fix and then add all test cases including the problematic one.

gbtami:
Yea, I just added praseo's position to let us see that your fix works. If we want more comprehensive test, it needs tons of lines.
Both is OK. Choose what you think is simpler for you.

ianfab:
Sorry for the confusion, I thought this was a PR against master.

PR #762, opened 2024-03-05
Musketeer variant

alperatalayn:


ianfab:
Thank you. Since there already are several other open PRs review will likely take some time. Could you please in the meantime try to fix the errors/warnings reported by CI?

ianfab:
@alperatalayn @cross-of-north @sayantandey @chris-heldeis Thanks for your PRs. May I ask you to align on your PRs and come up with a single PR? I have fairly limited time for reviews, so reviewing 4 versions of the same PR is not really practicable for me. Would be very nice if you get to a consensus on which is the most refined version that should be reviewed.

cross-of-north:
I consider my version to be the most refined.

- @alperatalayn initial version introduces basic structures to support musketeer variant , but breaks build and tests. Also it does not work.

- @chris-heldeis version fixes some problems, but some tests still fail (at least it was so last week, and new commits don't contain anything relevant). Also Python and JS binding versions are incremented; I can't say if it is useful or not.

- @sayantandey version: may be I'm not aware of some Github features, but in my opinion this version does not contain anything except @alperatalayn commits, so it's effectively empty.

- @cross-of-north (mine) version - build and tests are all green, contains improvements/fixes to make the variant actually playable and compatible with GUI (Winboard) and with implementations in other engines.

PR #763, opened 2024-03-06
wallOrMove bug fix

RainRat:
was able to trigger the asserts and fix them


ianfab:
Thanks, and sorry for the delay

PR #766, opened 2024-03-25
Add missing dragon promotion

gbtami:
See in https://github.com/gbtami/pychess-variants/issues/1338#issuecomment-2018334956


ianfab:
Thanks

PR #771, opened 2024-03-29
Fixes for PR #762 (Musketeer variant)

cross-of-north:
Fixed typos and other reasons why tests are failing in #762 .
Now all tests should be green.
Hope I haven't broke the original idea.


cross-of-north:
[replaced with #772 ]

PR #772, opened 2024-03-29
Fixes for PR #762 (Musketeer variant)

cross-of-north:
Fixed typos and other reasons why tests are failing in #762 .
Now all tests should be green.
Hope I haven't broke the original idea.

[replacement of #771]


musketeerchess:
Hi, 
i tested the executables from these changes and they perfectly work. 

Fabian, can you check them yourself and then merge with main code? 
Thanks

ianfab:
@cross-of-north Thanks. Please keep playing strength improvements and game logic in separate PRs, because the former is more optional and therefore more likely to be rejected if changes are too big and gains too specific to a single variant.

Also please note that changes removing existing features, e.g., replacing custom piece slots by built-in types, have very few chance of getting merged, since the benefits would need to significantly outweigh the downsides of the removed flexibility.

cross-of-north:
I suppose that there will be no more changes in this branch.
So it's ready for review and/or merge.

PR #795, opened 2024-05-15
Add missing rules for minihouse variant

DoTuanPhong:
missing rules added: 
1. Only promote to Rook, Knight, Bishop.
2. White to promote at 6th rank.
3. No double step.

### DIFF
--- a/src/variants.ini
+++ b/src/variants.ini
@@ -575,6 +575,9 @@ promotionPawnTypes = -
 maxRank = 6
 maxFile = 6
 startFen = 2bnrk/5p/6/6/P5/KRNB2
+promotionPieceTypes = nbr
+promotionRegionWhite = *6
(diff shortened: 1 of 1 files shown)


src/variants.ini
```diff
@@ -575,6 +575,9 @@ promotionPawnTypes = -
 maxRank = 6
 maxFile = 6
 startFen = 2bnrk/5p/6/6/P5/KRNB2
+promotionPieceTypes = nbr
+promotionRegionWhite = *6
+doubleStep = f
```
ianfab, line 580: Thanks for your PR. Technically anything other than `true` works here, but for consistency it would be good to change to `false`.

ianfab:
Closed via 81e3ceec0b16b76a6b2de7da92d15a2dcab362c3

PR #805, opened 2024-07-04
Priority drops

inversed-ru:
Implementation of priority drops, a feature I needed for my chess variant. Pieces from `priorityDropTypes` must be dropped before dropping any other pieces. In my variant, I use this feature to control the early game setup, even though I don't technically impose a separate setup phase: stronger pieces can be dropped only after all the weaker ones have been dropped.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -1086,6 +1086,8 @@ bool Position::legal(Move m) const {
           if (popcount((DarkSquares & to ? DarkSquares : ~DarkSquares) & pieces(us, BISHOP)) + 1 > (count_with_hand(us, BISHOP) + 1) / 2)
               return false;
   }
+  if (type_of(m) == DROP && (!var->isPriorityDrop[type_of(moved_piece(m))]) && priorityDropCountInHand[us] > 0)
+      return false;
 
   // No legal moves from target square
   if (immobility_illegal() && (type_of(m) == DROP || type_of(m) == NORMAL) && !(PseudoMoves[0][us][type_of(moved_piece(m))][to] & board_bb()))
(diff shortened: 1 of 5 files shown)


src/position.cpp
```diff
@@ -1086,6 +1086,8 @@ bool Position::legal(Move m) const {
           if (popcount((DarkSquares & to ? DarkSquares : ~DarkSquares) & pieces(us, BISHOP)) + 1 > (count_with_hand(us, BISHOP) + 1) / 2)
               return false;
   }
+  if (type_of(m) == DROP && (!var->isPriorityDrop[type_of(moved_piece(m))]) && priorityDropCountInHand[us] > 0)
```
isaacl, line 1089: For efficiency, I think these two new conditions should be checked in reverse order.

src/variant.h
```diff
@@ -90,6 +90,7 @@ struct Variant {
   bool mustCapture = false;
   bool mustDrop = false;
   PieceType mustDropType = ALL_PIECES;
+  bool isPriorityDrop[PIECE_TYPE_NB] = {};
```
ianfab, line 93: It would be more consistent and storage efficient to use a `PieceSet` here, similar as for properties. This would also simplify the parsing, since it can reuse existing logic. It could perhaps also already be made color-specific, similar as `promotionPieceTypes` or the like.

src/position.cpp
ianfab, line 1089: Since this is fairly trivial logic and does not depend on other moves, it should be more suitable to validate this in move generation and pseudo-legal validation instead of in legal validation.

alwey:
I think this should be implemented in the move generation of the individual variant.

PR #816, opened 2024-07-20
Do more full window searches

EarlyEdition:
Remove the `value < beta` condition for doing full window searches. 
As an added bonus the condition for full-window search is now much more similar to other fail-soft engines.

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -1368,10 +1368,9 @@ moves_loop: // When in check, search starts from here
           }
       }
 
-      // For PV nodes only, do a full PV search on the first move or after a fail
-      // high (in the latter case search only if value < beta), otherwise let the
-      // parent node fail low with value <= alpha and try another move.
-      if (PvNode && (moveCount == 1 || (value > alpha && (rootNode || value < beta))))
+      // For PV nodes only, do a full PV search on the first move or after a fail high,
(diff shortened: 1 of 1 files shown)


ianfab:
I do upstream merges in order, I am not interested in random cherry-picks out of order unless maybe for urgent bugs that can't wait for the other merges.

EarlyEdition:
> I do upstream merges in order

I think there is no need to merge all the upstream commits because they are specifically tested for **official-stockfish** and probably not useful for **fairy-stockfish**. I'd like to know @ddugovic's opinion on this PR.

> I am not interested in random cherry-picks out of order unless maybe for urgent bugs that can't wait for the other merges.

This is not random cherry-picks. This is a wonderful idea from @cj5716 which can help the search engines to do better. It does not matter whether it is **fairy-stockfish** or **official-stockfish**.

ianfab:
Thanks, but I don't need additional input on this. While I usually appreciate input on development, this is not a matter of development, this is a question of maintenance style. And I told you my decision as a maintainer and that is not subject to discussion, because the reasoning has nothing to do with the content of the PR. Please refrain from making unnecessary noise in this project, because I only have limited time. If you are just here out of frustration that official SF blocked you, then please let it be.

PR #827, opened 2024-09-27
Add sudoku chess variant

yusitnikov:
The PR introduces a new chess variant - "sudoku chess".
In a few words, it's normal chess, but player is not allowed to capture opponent's pieces while player's pieces break 8x8 sudoku rules (i.e. there are repeating pieces of the same type in a file, rank or 4x2 box). Full rules reference: https://gist.github.com/yusitnikov/e81a5e4df2fce6a854db8a2950fa5cfc

Benchmark results:
- "Nodes searched" value didn't change comparing to the official version (checked with NNUE enabled and disabled).
- Total time is around 2-3% slower than the official version, and the reason for it is allocating more memory in the `StateInfo` struct. I guess that it could be solved by allocating the memory dynamically only when the variant is enabled, but it will add complexity to the codebase. Is it worth optimizations?

Compatibility:
I personally plan using the variant only with classic setup: 8x8 board, standard initial position, normal chess rules + sudoku.
But theoretically, there's nothing in the variant implementation that prevents using it with:
- `LARGEBOARDS` compilation option.
- Boards of different sizes.
- Mixed with other chess variants.

Issues and potential improvements:
- Variant doesn't detect stalemates (because of replacing king with commoner).
- Position evaluation doesn't take variant specifics into account yet.
- `do_move` re-calculates the sudoku conflicts map from the scratch. It should be possible to optimize.

Sorry for a large PR! It's the smallest possible chunk of changes that will have a value for the variant support.

### DIFF
--- a/src/position.h
+++ b/src/position.h
@@ -59,6 +59,8 @@ struct StateInfo {
 
   // Not copied when making a move (will be recomputed anyhow)
   Key        key;
+  int        sudokuConflictsCount[COLOR_NB];
+  int        pieceCountInSudokuHouse[COLOR_NB][PIECE_TYPE_NB][SH_NB][FILE_NB];
   Bitboard   checkersBB;
   Piece      unpromotedCapturedPiece;
   Piece      unpromotedBycatch[SQUARE_NB];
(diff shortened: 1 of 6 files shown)


ianfab:
Thanks for your PR. In view of the many open feature addition PRs and the huge backlog of maintenance tasks (e.g. releases, upstream merges) I unfortunately have to say though that I am currently fairly hesitant to add features that are quite specific to rare/exotic variants, so I will defer the decision on this PR for quite some time until maintainability is improved. I hope you understand.

src/position.h
```diff
@@ -1500,6 +1520,35 @@ inline bool Position::allow_virtual_drop(Color c, PieceType pt) const {
...
 
+inline bool Position::sudoku_boxes() const {
+  assert(var != nullptr);
+  return var->sudoku && var->sudokuBoxWidth && var->sudokuBoxHeight;
+}
+
+inline int Position::sudoku_box_of(Square s) const {
+  assert(var != nullptr);
+  assert(sudoku_boxes());
+  return rank_of(s) / var->sudokuBoxHeight * (files() / var->sudokuBoxWidth) + file_of(s) / var->sudokuBoxWidth;
```
cab101, line 1531: you can use precalculated array instead of this, like `sudoku_box_of[to]`

yusitnikov: Right, but it's not a heavy calculation, and it depends on variant parameters, so I prefer not to over-complicate the code

PR #840, opened 2024-11-20
Implement fog_fen()

gbtami:
To prevent leaking FEN info in Fog of War games to the client side, the only correct solution is to create the "visible" FEN on the server side and send only that to the browser.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -681,7 +681,7 @@ Position& Position::set(const string& code, Color c, StateInfo* si) {
 /// Position::fen() returns a FEN representation of the position. In case of
 /// Chess960 the Shredder-FEN notation is used. This is mainly a debugging function.
 
-string Position::fen(bool sfen, bool showPromoted, int countStarted, std::string holdings) const {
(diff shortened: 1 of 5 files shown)


src/position.cpp
```diff
@@ -825,6 +825,41 @@ string Position::fen(bool sfen, bool showPromoted, int countStarted, std::string
...
+  // Everything else is invisible
+  fog = ~visible & board_bb();
+
+  // Fill in invisible squares with walls
+  while (fog)
+  {
+    Square sq = pop_lsb(fog);
+    if (piece_on(sq)) remove_piece(sq);
+
+    st->wallSquares |= sq;
```
ianfab, line 855: This method looks getter-like from its name, but seems to mess with the state here. Would be nicer to make this method `const` just like `fen()` itself.

gbtami: First I made it const, but because remove_piece itself is not const, compilation failed.

gbtami: Isn't that a problem that I have to replicate almost all of the code of `fen()`?

gbtami: Implemented your bitboard argument suggestion.

PR #846, opened 2024-12-11
chore(license): update license classifier to `GPL-3.0-or-later`

chenrui333:
Update to use SPDX license identifier, which also matches with the license header.

relates to https://github.com/Homebrew/homebrew-core/pull/200703/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ if "64bit" in platform.architecture():
 
 CLASSIFIERS = [
     "Development Status :: 3 - Alpha",
-    "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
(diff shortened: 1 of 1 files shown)


gbtami:
Fairy-Stockfish is derived from https://github.com/official-stockfish/Stockfish which is GPL v3.

ianfab:
I was not aware that this license identifier has been deprecated for its ambiguity (https://www.gnu.org/licenses/identify-licenses-clearly.html), so looks good to me. @gbtami Stockfish and hence also Fairy-Stockfish both have the "or later" clause in it.

chenrui333:
Thanks @ianfab!

PR #860, opened 2025-01-12
Adding LUA bindings

subvertnormality:
The Lua bindings provide a powerful interface to the chess variant engine. They require Lua 5.1 for LuaJIT compatibility. See tests/lua/README.md for installation and usage instructions.


ianfab:
Thanks for the PR. There definitely is some value in having additional language bindings, but I currently do not see a concrete use case that would justify the maintenance overhead this creates, especially because it introduces git submodules, and also given my lack of familiarity with Lua. So for now I am very hesistant about incorporating it.

subvertnormality:
Fair enough - I plan to use them in a project I’m working on and opened the PR in case others might find them valuable. No problem if they’re not merged. They’ll be here if it turns out you’d prefer it to be part of main-line.

PR #893, opened 2025-07-27
Fix Lack of Pieces in Hand in Flipello Variants

PraseodymiumSpike:
60 per side for Othello
64 per side for Reversi
96 per side for 10-by-10 Othello


PraseodymiumSpike:
to fix https://github.com/fairy-stockfish/Fairy-Stockfish/issues/504

ianfab:
Thanks. Reversi should however be reverted to 32 since players actually get 32 disks each, no unlimited supply.

PraseodymiumSpike:
I reverted it to 32.

PR #906, opened 2025-08-20
Fix bug for drawn KQk positions being evaluated as winning in atomic

johndoknjas:
For some reason, pruning causes stockfish to misevaluate this endgame. E.g., for `2q1k3/4K3/8/8/8/8/8/8 w - - 0 1`, the evaluation is a draw at low depths, but then winning at higher ones.

May or may not be related, but `pos.checkers()` doesn't apply for atomic. Maybe this matters somehow for the pruning logic? Also, if quiet moves are pruned for the attacker, maybe cases where the attacking king should move away are pruned?

As an aside, this PR also updates `endgame.cpp`, but it isn't needed for fixing the KQk bug. I noticed in the function it's assumed it won't be called by search if the weaker side is in check, but this actually isn't the case. I've updated the function accordingly. Seems like the only real effect though is that running `eval` for FENs like `8/8/8/8/4k3/8/6q1/6K1 w - - 0 1` and `8/8/8/8/4k3/5q2/6K1/8 w - - 0 1` will now give the correct evaluation. Search is able to recover from wrong evals being given.

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -1141,7 +1141,7 @@ moves_loop: // When in check, search starts from here
       if (  !rootNode
           && (pos.non_pawn_material(us) || pos.count<ALL_PIECES>(us) == pos.count<PAWN>(us))
           && bestValue > VALUE_TB_LOSS_IN_MAX_PLY
-          && (!pos.blast_on_capture() || pos.count<ALL_PIECES>() > 3))
+          && (Options["UCI_Variant"] != "atomic" || pos.count<ALL_PIECES>() > 3 || pos.count<QUEEN>() != 1))
       {
           // Skip quiet moves if movecount exceeds our FutilityMoveCount threshold
           moveCountPruning = moveCount >= futility_move_count(improving, depth, pos);


ianfab:
Thanks! The KQK indeed is a bit buggy.

Regarding the pruning, that change looks quite artificial to me, so I tried to debug it a bit, and it seems to me like there is a bug in SEE for atomic when both kings are exploded in a capture by another piece, i.e., here the queen. In that case it is not very well defined whether it is considered a good or bad move by SEE since that randomly depends on which king is found first in this loop:
https://github.com/fairy-stockfish/Fairy-Stockfish/blob/4d1d0cc8fd70711dd67788168306540c8d2ba037/src/position.cpp#L2418-L2426

I am currently testing a fix for this buggy atomic SEE logic, which seems to at least massively reduce this issue, so would fit my suspicion. That would also explain why your fix works, because it prevents the buggy SEE pruning to be executed in the first place.

Therefore for now I won't merge the fix as is, since it seems now like this is just a workaround, and instead fixing SEE seems like the right way.

Either way, regardless of whether it will be merged, thanks a lot for investigating the issue and pointing towards the solution.

johndoknjas:
@ianfab Sounds good! Addressing the root issue is definitely preferable. I had tried to do this myself but was unsuccessful.

ianfab:
I pushed my fix suggestion to #907

PR #909, opened 2025-08-22
Fix bug with antichess K vs N

johndoknjas:
- In the first if statement of the diff, `if (pos.side_to_move() == strongSide && (attacks_bb<KNIGHT>(NSq) & KSq))` is a bug. I think the intention is to see if the king is attacking the knight, so I fixed it to that.
- In some cases where the king is to move and it's on the edge (e.g., Kg1 and Ne2), the knight attacking the king actually does mean that the knight wins (but only if the king is not attacking any other squares the knight attacks). So I rewrote the code to handle all cases where the king and knight attack each other or share attacks.

Briefly tested with a few positions, but haven't done a fishtest (http://www.variantfishtest.org:6543/tests doesn't load for me).

### DIFF
--- a/src/endgame.cpp
+++ b/src/endgame.cpp
@@ -1033,7 +1033,7 @@ Value Endgame<RK, EG_EVAL_ANTI>::operator()(const Position& pos) const {
 }
 
 
-/// K vs N. The king usally wins, but there are a few exceptions.
+/// K vs N. The king usually wins, but there are a few exceptions.
 template<>
 Value Endgame<KN, EG_EVAL_ANTI>::operator()(const Position& pos) const {
 
@@ -1045,11 +1045,13 @@ Value Endgame<KN, EG_EVAL_ANTI>::operator()(const Position& pos) const {
   Bitboard knightAttacks = attacks_bb<KNIGHT>(NSq);
   bool strongSideToMove = pos.side_to_move() == strongSide;
 
-  if (strongSideToMove ? kingAttacks & NSq : knightAttacks & KSq) return mated_in(1);
-  if (kingAttacks & knightAttacks) return mate_in(2);
-  if (strongSideToMove ? knightAttacks & KSq : kingAttacks & NSq) return mated_in(3);
+  if (strongSideToMove ? kingAttacks & NSq : knightAttacks & KSq) return VALUE_TB_LOSS_IN_MAX_PLY + 1;
+  if (kingAttacks & knightAttacks) return VALUE_TB_WIN_IN_MAX_PLY - 2;
+  if (strongSideToMove ? knightAttacks & KSq : kingAttacks & NSq) return VALUE_TB_LOSS_IN_MAX_PLY + 3;
 
-  Value result = VALUE_KNOWN_WIN + push_to_edge(NSq, pos) - push_to_edge(KSq, pos);
+  Value result = Value(push_to_edge(NSq, pos)) - push_to_edge(KSq, pos);
+  if (!(KSq & (SQ_A1 | SQ_A8 | SQ_H1 | SQ_H8)))
+      result += VALUE_KNOWN_WIN;
 
   return strongSideToMove ? result : -result;
 }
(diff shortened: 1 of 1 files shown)


ianfab:
I read this code again and as far as I can see it was fully intentional.

Scenario 1: If the knight attacks the king and the king is to move, the knight side always wins by premoving to the king square.
Scenario 2: If the king and knight attack a common square and the knight is to move, it can sacrifice itself on that square.

Scenario 2 seems to miss the scenario where the knight already attacks the king, apart from that it looks fine. And in case of mandatory captures pruning is generally skipped, so it is unlikely to be a problem.

Note that you can not use mate values in endgame evaluation, because it does not know about the current depth at which this position occurred (ss->ply).

Regarding fishtest, I put it down some time ago due to inactivity to save costs, and it likely won't come back.

ianfab:
For me the example you mentioned works fine with your current code. However, I noticed that it is much less trivial than I thought to figure out all exceptions where the knight wins than I had thought, and now better understood why you initially wrote the code the way you did. So for testing I removed the mate scores and added an exception for king in the corner, because e.g. with king in the corner and knight in the corresponding center square (e.g., Kh1 Ne4) the knight wins with either(!) side to move.

I have the feeling the king in the corner exception might be a bit overkill, but didn't have time to do any extensive comparison in terms of how it helps search yet, but at least in terms of evaluation this seems pretty accurate.

```
/// K vs N. The king usally wins, but there are a few exceptions.
template<>
Value Endgame<KN, EG_EVAL_ANTI>::operator()(const Position& pos) const {

  assert(pos.endgame_eval() == EG_EVAL_ANTI);

  Square KSq = pos.square<COMMONER>(strongSide);
  Square NSq = pos.square<KNIGHT>(weakSide);
  Bitboard kingAttacks = attacks_bb<KING>(KSq);
  Bitboard knightAttacks = attacks_bb<KNIGHT>(NSq);
  bool strongSideToMove = pos.side_to_move() == strongSide;

  if (strongSideToMove ? kingAttacks & NSq : knightAttacks & KSq)
      return VALUE_TB_LOSS_IN_MAX_PLY + 1;
  if (kingAttacks & knightAttacks)
      return VALUE_TB_WIN_IN_MAX_PLY - 2;
  if (strongSideToMove ? knightAttacks & KSq : kingAttacks & NSq)
      return VALUE_TB_LOSS_IN_MAX_PLY + 3;

  Value result = Value(push_to_edge(NSq, pos)) - push_to_edge(KSq, pos);
  if (!(KSq & (SQ_A1 | SQ_A8 | SQ_H1 | SQ_H8)))
      result += VALUE_KNOWN_WIN;

  return strongSide == pos.side_to_move() ? result : -result;
}
```

johndoknjas:
@ianfab I've applied your patch, thanks. There's also scenarios where a knight can beat a king anywhere on the edge. E.g., Ke1 and Nc2/Ng2, when it is the king to move. This also means any position with Ke1 and the knight a move away from c2/g2 will be a win for the knight, when it is its turn. I don't know whether or not it's worth covering these cases though.

For what I was talking about for the longer mate, this branch finds a mate in 8 by depth 18, and a mate in 7 by depth 21. For the master branch though, a mate in 7 is found by depth 15.

ianfab:
Thanks, right, I forgot about the other cases. However, there is a trade-off between complexity and accuracy, so I would like to stop before those non-trivial exceptions, as mentioned my own exclusion of corner kings already almost felt like overkill, although it at least codewise is simple. There is still search to handle the rest, it is sufficient to be a fairly decent heuristic. If we really want to make sure to avoid false win predictions, something simpler could be to just generalize the exception I made, like
```
-  if (!(KSq & (SQ_A1 | SQ_A8 | SQ_H1 | SQ_H8)))
+  if (!(KSq & (FileABB | FileHBB | Rank1BB | Rank8BB)))
```
I think there isn't any case with a king not on the edge that doesn't win (apart from the conditions captured before already), so that would keep it conservative and still fairly simple.


As a side note, I think we are using different testing conditions (I use `./stockfish bench antichess 16 1 25 test.epd` with the FEN `n7/8/8/8/8/8/8/7K w - - 0 1` in the test.epd), as for me the latest version finds mate later than the commit before. Maybe you use 8x8 while I currently use the largeboard version. Either way, the difference is small.

PR #918, opened 2025-09-19
Add Camelrider, Dabbabarider, Alfilrider and En passant for any piece type

MoKy1978:


src/movegen.cpp
```diff
@@ -299,15 +298,16 @@ namespace {
 
         Bitboard attacks = pos.attacks_from(Us, Pt, from);
         Bitboard quiets = pos.moves_from(Us, Pt, from);
-        Bitboard b = (  (attacks & pos.pieces())
-                       | (quiets & ~pos.pieces()));
-        Bitboard b1 = b & target;
+        Bitboard b = ((attacks & pos.pieces()) | (quiets & ~pos.pieces()));
+
+        Bitboard epSquares = pos.variant()->enPassantTypes[Us] & Pt ? attacks & pos.ep_squares() & ~pos.pieces() : Bitboard(0);
```
ianfab, line 303: Please revert the changes in this file for this PR. Regardless of whether you consider it a bugfix or a change of specification (as I do), it is independent of the rest of the PR, so it should a separate PR to avoid mixing discussions.

ianfab (commented):
Thanks for your PR. Since I plan to make a release soon and don't want to introduce any risk before that, I won't merge it for the next weeks. However, in principle adding more movements is useful, so I will have a closer look at a later point. One thing worth checking either way is to compare speed (bench) before and after to see if it has any significant impact on performance. I assume it won't, but worth checking.

src/movegen.cpp
MoKy1978: OK, but first I need to figure out how to do it

PR #928, opened 2025-10-18
Added the Benedict Morph variant

Natfay:
Implements the "Benedict Morph" chess variant; a morph-on-capture rule inspired by the original Benedict game, but simplified for clarity and engine compatibility.

Rules summary
----------------
- When a piece captures an opponent's piece, it *morphs* into the type of the captured piece (keeping its original color).
    • Example: a white bishop capturing a black knight becomes a white knight on the destination square.
- The captured piece is removed as usual.
- All other rules of orthodox chess apply (castling, en passant, promotions, check/mate conditions, etc.).
- The king is **rex-exclusive**:
    • The king can capture, but never morphs, it always remains a king.

Implementation details
-------------------------
- Added new variant factory `benedictmorph_variant()` in variant.cpp.
- Added four fields in `StateInfo` to record morph data: `didMorph`, `morphedFrom`, `morphedTo`, `morphSquare`.
- Modified `Position::do_move()`:
    • Detect capture and perform morph before gating.
    • Skip morph if the mover is a king (rex-exclusive).
    • Update NNUE dirtyPiece list to reflect the final piece identity.
- Modified `Position::undo_move()` to reverse the morph.

Rex-exclusive by default. Future extensions could include a non-rex-exclusive mode or a "no king captures" variant.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -1580,7 +1580,7 @@ void Position::do_move(Move m, StateInfo& newSt, bool givesCheck) {
       captured = NO_PIECE;
   }
   st->capturedpromoted = is_promoted(to);
-  st->unpromotedCapturedPiece = captured ? unpromoted_piece_on(to) : NO_PIsECE;
+  st->unpromotedCapturedPiece = captured ? unpromoted_piece_on(to) : NO_PIECE;
   st->pass = is_pass(m);
 
   assert(color_of(pc) == us);


ianfab:
Thanks for the PR.

From a first glance the changes generally look good to me. However, at the current moment in a phase before a release, most PRs other than bugfixes and trivial changes are on hold. Also despite the changes being fairly isolated, the feature is also very niche, so I will need to decide on the maintenance trade-off. I will revisit it at a later point when there is a more suitable moment of a release cycle.

PR #951, opened 2025-12-25
Fix petrify, add self-capture and last piece capture

MoKy1978:


MoKy1978:
selfCapture: allow capturing own pieces (except king)
capturingLast: win by capturing opponent's last piece
petrify: petrified pieces no longer go to hand

PR #1006, opened 2026-06-12
Feature/xiangqi chinese notation

gbtami:


gbtami:
@ianfab when codex implemented Tori Shogi Japanese notation in https://github.com/fairy-stockfish/Fairy-Stockfish/pull/1006/changes/5809b2868483d985403c8515cf1f8790c8cd5673 I noticed that `Position` class `Variant` info has no variant name info at all, so he did almost full variant rule check to identify Tori (and a weird one for Dobutsu as well). Should I use startFen check instead? I mean I'm not sure adding a new variant name as string to Variant would be a good idea or not.

ianfab:
I think it should perhaps be more declarative, or at least conceptually more clearly defined. To me conceptually there is a notation system, piece names/characters, and language specific piece names/characters. The first we have an enum for, the second is in the pieceToChar, and the third is currently kind of somewhat folded into the notation system, because sometimes in a different language not just the characters are exchanged, but the notation system differs further.

One idea could be to consider piece translations a variant specific attribute, similar as the pieceToChar is the English character map, there could be other language specific ones. Not sure if I entirely like it, because it could get a lot, but at least conceptually clean. If we consider language specific notation out of scope, we might consider them separate notation systems. Or we have some kind of enumerated notation subsystems that define character maps and could potentially also be a variant property.

gbtami:
Codex implemented a more declarative solution for localized notation in https://github.com/fairy-stockfish/Fairy-Stockfish/pull/1006/changes/0769c848706d3c4db21a3e579c834ae6d28f9e03

PR #1012, opened 2026-07-18
Update Dragonfly variant rules in variants.ini

gbtami:
dropNoDoubledCount = 0 causes to prevent pawn drops, so the original rule to not allow captured pawns to be dropped is enforced. Also to enable castling adding castling rights to the startFen is fixed.


PraseodymiumSpike:
brilliant workaround

PR #1018, opened 2026-07-23
Test Capahouse moves with populated pockets

jlmalone:
## What changed

Add a large-board perft regression for a Capahouse position with both pockets populated.

## Why

The existing Capablanca perft covers ordinary 10x8 move generation, while the current pocket tests cover 8x8 variants. This position exercises Capablanca compound-piece drops, pocket ownership, and the much larger legal move set produced by populated Capahouse reserves.

The expected depth-two count was independently reproduced with a separate Capablanca move generator.

## Validation

- `make -j2 ARCH=apple-silicon largeboards=yes all=yes build`
- `./stockfish check variants.ini`
- `../tests/perft.sh largeboard`
- `../tests/perft.sh all`

### DIFF
--- a/tests/perft.sh
+++ b/tests/perft.sh
@@ -166,6 +166,7 @@ if [[ $1 == "all" ||  $1 == "largeboard" ]]; then
   expect perft.exp shoshogi startpos 4 445372 > /dev/null  # configurable pieces
   expect perft.exp yarishogi startpos 4 158404 > /dev/null  # configurable pieces
   expect perft.exp capablanca startpos 4 805128 > /dev/null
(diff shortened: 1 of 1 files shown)


ianfab:
Thanks for your PRs. Please use the draft status to indicate whether the PRs are supposed to be reviewed already or not.

Regarding this PR, it is a good point that while the large board variant perfts do have drop variants they don't have pieces in hand, so drops won't appear much in these tests. Makes sense.

jlmalone:
Yes, these PRs are intentionally drafts until they are ready for maintainer review. This perft coverage is ready for review; I will mark it ready once the related contribution set is settled.

ianfab:
Is there a human in the loop? If so, please let me know if all your PRs are considered ready from your side and set them accordingly.

jlmalone:
I believe they are ready yes.

PR #1021, opened 2026-07-29
Count checks against pseudo-royal pieces

### LINKED ISSUE #975: Check counting does not work for extinction pieces with extinctionPseudoRoyal = true
When extinctionPsuedoRoyal = true, the last remaining extinction piece should be treated as a royal piece, meaning it should be able to be checked. However, when checkCounting = true, the number of checks does not decrease when the last remaining extinction piece is checked. This also applies to duple checks when dupleCheck = true.

### Steps to reproduce

Create a variant with the following parameters:

```
extinctionValue = loss
extinctionPieceTypes = (insert desired piece types here)
extinctionPseudoRoyal = true

checkCounting = true
```

Then, set the position such that there is only one extinction piece left for black and check it with white. Observe that the number of checks does not decrease.

Here's an example variant:

```
[test_extinction_check_count]

knight = n
queen = q
king = -

castling = false

extinctionValue = loss
extinctionPieceTypes = *
extinctionPseudoRoyal = true

checkCounting = true

startFen = 4n3/8/8/8/8/8/8/3Q4 w - - 9+9 0 1
```

Qe1 results in a check but the check counts remain as 9+9

jlmalone:
Fixes #975.

Check counting now recognizes checks against extinction pseudo-royal pieces after the resulting board state is available. The counter is incorporated into the resulting position key after pseudo-royal detection is refreshed.

The regression covers both a single pseudo-royal and a duple-check position. In each case, `Qe1` changes the check counter from `9+9` to `8+9`.

Validation:
- `make -j2 ARCH=apple-silicon build`
- `../tests/perft.sh chess`
- focused Expect regression passed directly
- `./stockfish bench`

The repository protocol wrapper could not run end-to-end on this macOS host because GNU `timeout` is not installed.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -1593,9 +1593,6 @@ void Position::do_move(Move m, StateInfo& newSt, bool givesCheck) {
   assert(captured == NO_PIECE || color_of(captured) == (type_of(m) != CASTLING ? them : us));
   assert(type_of(captured) != KING);
 
-  if (check_counting() && givesCheck)
-      k ^= Zobrist::checks[us][st->checksRemaining[us]] ^ Zobrist::checks[us][--(st->checksRemaining[us])];
-
   if (type_of(m) == CASTLING)
   {
       assert(type_of(pc) != NO_PIECE_TYPE);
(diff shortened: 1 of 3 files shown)


ianfab:
The code change itself looks fine to me, but the test implementation does not align well with test conventions in this repo. Such kind of tests usually are implemented via the bindings, e.g., see test.py.

PR #1024, opened 2026-08-14
Package pyffish type information per PEP 561

gbtami:
## Motivation

  `pyffish.pyi` was previously installed through:

  ```python
  data_files=[("", ["pyffish.pyi"])]

  This does not place the stub next to the compiled pyffish extension in site-packages. For wheels, it is stored under the wheel’s .data/data directory and can end up at the virtual
  environment root after installation. As a result, type checkers such as mypy do not discover it.

  Additionally, PEP 561 does not define typed single-file/module-only distributions as the recommended layout. Typed libraries should expose their type information from a package
  containing a py.typed marker.

  ## Changes

  - Convert pyffish from a top-level extension module into a package.
  - Install the native extension privately as pyffish._pyffish.
  - Re-export its public API from pyffish/__init__.py, preserving the existing import pyffish API.
  - Move the stub to pyffish/__init__.pyi.
  - Add the required pyffish/py.typed marker.
  - Include the typing files in wheels and source distributions through package_data and MANIFEST.in.
  - Add the missing FEN_INVALID_PROMOTED_PIECE constant to the stub.
  - Add a regression test that verifies the typing metadata is installed.
  - Adjust the cibuildwheel test command so the source package does not shadow the installed wheel.

  ## Compatibility

  The public Python API remains unchanged:

  import pyffish

  pyffish.version()
  pyffish.legal_moves(...)

  Only the internal compiled-module location changes from pyffish to pyffish._pyffish.

  ## Verification

  Tested using a clean Python 3.14 build and virtual environment:

  - Wheel contains pyffish/__init__.pyi, pyffish/py.typed, and the private native extension.
  - Source distribution contains all required package and typing files.
  - All 23 runtime tests pass.
  - mypy --strict discovers the installed types without a project-local stub.
  - python -m mypy.stubtest pyffish passes with no runtime/stub mismatches.

(co-authored with codex)


ianfab:
Thanks. I am not too deep into this, but as I understand this is the normal way to do it, so sounds good to me.

PR #1025, opened 2026-08-14
Package pyffish-alice type information per PEP 561

gbtami:


gbtami:
I accidentally created this PR. I can maintain pyffish-alice in my own fork, sry.

PR #1028, opened 2026-08-16
Treat SEE as neutral in Misere

ck-zhang:
SEE is based on the idea that losing material is bad, which is not true in Misere.

This patch treats SEE as neutral in Misere.

## Example

At 2,000 nodes with the Misere NNUE:

```text
4Q2b/8/5k2/8/8/8/P7/KB4R1 w - - 0 1
```

Before the patch, Fairy chooses `g1g2` and the game draws. With the patch it finds `g1g7`, reports `mate 2`, and forces its own checkmate.

## Results

- 512-game tactical validation: 118 wins, 342 draws, 52 losses (`+44.95` Elo-like; 95% CI `+33.97` to `+56.71`).
- 2,048-game broad screen: 9 wins, 2,034 draws, 5 losses (`+0.68`; confidence interval crosses zero).

This patch fixes a real forced-selfmate blind spot.
No changes for non-misère variants.

### DIFF
--- a/src/position.cpp
+++ b/src/position.cpp
@@ -2492,7 +2492,9 @@ bool Position::see_ge(Move m, Value threshold) const {
       return extinction_value() < VALUE_ZERO;
 
   // Do not evaluate SEE if value would be unreliable
-  if (must_capture() || !checking_permitted() || is_gating(m) || count<CLOBBER_PIECE>() == count<ALL_PIECES>())
+  if (   endgame_eval() == EG_EVAL_MISERE
+      || must_capture() || !checking_permitted() || is_gating(m)
+      || count<CLOBBER_PIECE>() == count<ALL_PIECES>())
       return VALUE_ZERO >= threshold;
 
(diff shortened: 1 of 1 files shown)


ianfab:
Thanks, it is a good idea to address the lower relevance of SEE in misere.

Generally I like to keep variant specific changes minimal in order to keep the code manageable. Therefore, one thing I did in the past with SEE specifically is that for variants where SEE is not a very useful metric, I simply collapse it to only check if the threshold is above or below 0. In most cases this has a similar effect as skipping usage of SEE.
https://github.com/fairy-stockfish/Fairy-Stockfish/blob/c19b5f6c66894fdb0e88d0dd100e3885f744760a/src/position.cpp#L2494-L2496

This would be my preferred way to implement such an idea for misere if it works similarly well as the targeted search changes.

ck-zhang:
Thanks for the feedback! I tested it locally and got similar results. Really excited to contribute to such a cool C++ codebase for the first time :)

ianfab:
Thanks!

Issue #966, opened 2026-01-08
Black custom pieces have no value


HGMuller:
The PieceValue array in psqt.cpp is initially filled with VALUE_ZERO, both for white and black pieces. But the loop in the init routine there only runs over piece types extracted from a PieceSet, i.e. uncolored types:

  for (PieceSet ps = v->pieceTypes; ps;)
  {
      PieceType pt = pop_lsb(ps);
      if (is_custom(pt))
      {
          PieceValue[MG][pt] = piece_value(MG, pt);
          PieceValue[EG][pt] = piece_value(EG, pt);
      }

      if (PieceValue[MG][pt] > PieceValue[MG][strongestPiece])
          strongestPiece = pt;
  }

The loop that adapts values then runs from PAWN to KING, i.e. also for white pieces only. The CapturePieceValue and EvalPieceValue are also assigned to ~pc, but this is not done for PieceValue itself. (Which is used for updating nonPawnMaterial.

When I put an assert(st->nonPawnMaterial[BLACK] > 0) at the end of do_move, it triggers in my test variant with only custom pieces before FSF makes a single move.

PR #1033, opened 2026-08-30
bent riders

fhoudebert:
# Add the Griffon, a bent rider

Adds the Griffon as a predefined piece type: one diagonal step, on which it may
stop, then an unlimited orthogonal ride away from the starting square, over
empty squares only. Movement follows Jocly's `cbGriffonGraph` (a `cbSkiGraph`
over the four diagonal steps with a 45° bend, unlimited range, corner square
included as a destination).

Bent riders are currently not expressible in Fairy-Stockfish at all: `PieceInfo`
only has `steps`, `slider` and `hopper`, none of which can encode a change of
direction mid-move. This is the narrow version of the work @HGMuller did on the
`pieces` branch in #962 — Griffon only, no parser changes.

Refs #962, #966, #967.

## Why bent riders can't reuse the fast paths

Griffon reachability is symmetric (reachable iff both coordinate offsets are
non-zero and at least one is 1), but **the path is not**. A Griffon on a1
reaches b5 via b2-b3-b4; a Griffon on b5 reaches a1 via a4-a3-a2. The blocking
sets are disjoint.

That invalidates two assumptions: `checkSquares[pt]` is computed from the king
square and assumes reversibility, and `BetweenBB` is built as the intersection
of the two reciprocal attack sets, which for a1/b5 yields `{c2}` instead of the
real interposition squares `{a2, a3, a4}`.

Putting `RIDER_GRIFFON` in `NON_SLIDING_RIDERS` and `ASYMMETRICAL_RIDERS` is
enough to route around both, and no other correctness machinery was needed.
This matches what @HGMuller reported in #962: declaring the new rider types
non-sliding and asymmetrical was what made his king captures go away.

* `NON_SLIDING_RIDERS` makes `generate_all<EVASIONS>` fall back to
  `target = ~pos.pieces(Us)` and makes `legal()` skip its `between_bb`
  shortcut. The full `attackers_to()` at the end of `legal()` then covers pins.
* `ASYMMETRICAL_RIDERS` makes `slider_blockers()`, `attackers_to()` and
  `gives_check()` recompute attacks from each candidate square rather than
  reversing them.

## Changes

**types.h** — `GRIFFON` after `CENTAUR`. `RIDER_GRIFFON = 1 << 14`, placed after
every magic-backed rider, plus a `MAGIC_RIDERS` mask: `rider_attacks_bb`
indexes `magics[]` by bit position, so a rider bit without a magic table has to
be stripped before that loop.

**piece.h / piece.cpp** — a `bent[2][MOVE_MODALITY_NB]` map in `PieceInfo`,
keyed by the direction of the first leg, and `griffon_piece()` built like
`janggi_elephant_piece()`: `from_betza("F")` plus the four diagonal bent legs,
with `betza = "FyafsF"` for XBoard. Leaving the corner step in `steps` rather
than folding it into `bent` keeps "may the Griffon stop after one diagonal
step" a definition choice instead of hardcoding it.

**bitboard.h / bitboard.cpp** — no new magic table. Both continuation
directions of a diagonal leg are orthogonal, so one existing rook lookup per
corner square covers them, masked by a new `RayBB[8][SQUARE_NB]` (8 KB, 16 KB
with `LARGEBOARDS`) that discards the half pointing back at the origin. A
monolithic Griffon mask would span two files and two ranks, which looked
unwise on 12x10. `griffon_path_bb()` supplies the directed
`between_bb(s1, s2, GRIFFON)`.

**psqt.cpp** — `isSlider` and the board-size scaling only read `slider` and
`hopper`, so a piece whose movement lives in `bent` was silently classified as
a non-slider. Each bent leg is counted as 1.5 slider directions, putting the
Griffon at 2155 mg against 2590 for the queen (83%). Empty-board mobility is
21.4 squares against the queen's 22.75, and Jocly values them 8.3 vs 9.5 (87%),
so this is in the right region, but self-play tuning is the real answer.

**variants.ini** — documentation plus a `griffonchess` example variant.

## Testing

**No regression on standard chess.** `bench` gives 6180480 nodes both with and
without the patch. Perft from the start position to depth 6 = 119060324 and
Kiwipete to depth 5 = 193690690, both matching the known values.

**Move generation** cross-checked against an independent path-walking reference
over 400 random positions with friendly and enemy blockers, using `go perft 1`
on a variant whose blockers are `immobile` pieces so the move list contains only
Griffon moves: 0 disagreements.

**Check detection.** Two positions in `griffonchess`, same pieces, blocker moved
from b3 to a3:

| FEN | Expected | Result |
|---|---|---|

A symmetric implementation gets both backwards. Worth keeping as a regression
test.

**Stress.** 108,759 plies of random games in `griffonchess` on a `debug=yes`
build (300 games, 67 reaching a terminal position), plus `go depth 22` on K+G vs
K endgames, which is where #962 crashed. No assertion failures. That is
consistent with those crashes having been the #966 bug fixed by #967.

**Perft, `griffonchess` start position:** 20, 400, 8682, 187940, 4437378. These
come from this implementation, so they are a reproducibility reference rather
than ground truth. Cross-checking them against the Interactive Diagram would be
worth more than anything I can produce here.

## Cost

Comparing like with like — both as user-defined variants, so both off the
`fastAttacks` path — `griffonchess` runs at 336 knps against 581 knps for the
same variant with a queen, about 40% slower. That is the asymmetrical and
non-sliding paths plus 8 magic lookups per attack query, and it only affects
variants that use the piece.

Custom piece headroom drops from 26 slots to 25. `customPiece25` still parses,
so nothing user-visible breaks.

## Limitations

* Predefined piece only. `bent` is not reachable from Betza, so `customPieceN`
  cannot produce a Griffon.
* Only the full set of four diagonal legs is supported; partial sets and range
  limits would need per piece type direction tables. There is an assert for it.
* `checkSquares` stays the approximate forward set, exactly as it already is for
  `HORSE`, which is also in `ASYMMETRICAL_RIDERS`. It only feeds `QUIET_CHECKS`
  generation, so the cost is a little search quality, not correctness.
* NNUE does not know the piece, so Griffon variants run on classical eval.

## Questions before this is worth polishing

1. Do you want this at all as a standalone piece, or would you rather upstream
   @HGMuller's `pieces` branch now that #967 has landed? That is a bigger but
   far more general change, and this PR should be dropped in its favour.
2. Should the XBoard notation match his branch exactly (`yafsF`, versus
   `FyafsF` here, since Jocly's Griffon can stop on the corner square)?
3. Is the ~40% slowdown acceptable, or would you want per-quadrant magics up
   front?
4. Extend the parser in this PR so `customPieceN` can express bent riders? It
   would also give the Rhino/Manticore almost for free — same code path,
   `BishopMagics` instead of the rook ones.
5. Should I replace the approximate `checkSquares` with a proper reverse
   computation, or is matching the existing `HORSE` treatment fine?

bench: 6180480

### DIFF
--- a/src/bitboard.cpp
+++ b/src/bitboard.cpp
@@ -37,6 +37,7 @@ Bitboard PseudoMoves[2][COLOR_NB][PIECE_TYPE_NB][SQUARE_NB];
 Bitboard LeaperAttacks[COLOR_NB][PIECE_TYPE_NB][SQUARE_NB];
 Bitboard LeaperMoves[2][COLOR_NB][PIECE_TYPE_NB][SQUARE_NB];
 Bitboard BoardSizeBB[FILE_NB][RANK_NB];
+Bitboard RayBB[8][SQUARE_NB];
 RiderType AttackRiderTypes[PIECE_TYPE_NB];
 RiderType MoveRiderTypes[2][PIECE_TYPE_NB];
 
@@ -107,7 +108,9 @@ namespace {
                                                             {EAST  + 2 * SOUTH_EAST, 0}, {SOUTH + 2 * SOUTH_EAST, 0},
                                                             {SOUTH + 2 * SOUTH_WEST, 0}, {WEST  + 2 * SOUTH_WEST, 0},
                                                             {WEST  + 2 * NORTH_WEST, 0}, {NORTH + 2 * NORTH_WEST, 0} };
-  const std::map<Direction, int> GrasshopperDirectionsV { {NORTH, 1}, {SOUTH, 1}};
+  // Bent riders: the key is the direction of the first (single step) leg.
+const std::map<Direction, int> GriffonDirections { {NORTH_EAST, 0}, {SOUTH_EAST, 0}, {SOUTH_WEST, 0}, {NORTH_WEST, 0} };
+const std::map<Direction, int> GrasshopperDirectionsV { {NORTH, 1}, {SOUTH, 1}};
   const std::map<Direction, int> GrasshopperDirectionsH { {EAST, 1}, {WEST, 1} };
   const std::map<Direction, int> GrasshopperDirectionsD { {NORTH_EAST, 1}, {SOUTH_EAST, 1}, {SOUTH_WEST, 1}, {NORTH_WEST, 1} };
 
@@ -186,7 +189,21 @@ namespace {
     return b;
   }
 
-  Bitboard lame_leaper_attack(std::map<Direction, int> directions, Square s, Bitboard occupied) {
+  void init_rays() {
+
+  for (int i = 0; i < 8; ++i)
+      for (Square s = SQ_A1; s <= SQ_MAX; ++s)
+      {
+          Bitboard b = 0;
+          for (Square t = s + QueenDirections[i];
+               is_ok(t) && distance(t, t - QueenDirections[i]) == 1;
+               t += QueenDirections[i])
+              b |= t;
+          RayBB[i][s] = b;
+      }
+}
+
+Bitboard lame_leaper_attack(std::map<Direction, int> directions, Square s, Bitboard occupied) {
     Bitboard b = 0;
     for (const auto& i : directions)
     {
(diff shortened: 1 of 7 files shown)


fhoudebert:
The Betza parser needs to be extended so that bent riders can be declared as customPieceN.

HGMuller:
I have not examined your code, but you mention 8 lookups. In my 'pieces' fork I did not use 8 magic lookups for the Griffon, but 4. This could be done because the Griffon moves are the same as those of two vR on the sW squares, plus two sR on the vW squares, and both the vR and sR magic tables are already present.

Unfortunately the same did not work for the Manticore. Because the Bishops are not treated as seperate diagonal and anti-diagonal magics, and the pairs of legs that are on the same ray include all squares on that ray, while Bishop tables always have the Bishop square itself as a don't care, masking it out. So I did add new magic tables for diagonal and for anti-diagonal moves that contain all squares on those rays. Unfortunately the tables are pretty large. In principle these tables should also be usable for an Osprey, with a different mask (ignoring the square on the ray that the Osprey does not reach). The Manticore could have been done with 3 lookups, but that would have required very special tables, which could not be generated by the available routines, and would not be usable for an Osprey.

fhoudebert:
I finally added the Osprey to this pull request:
https://github.com/fairy-stockfish/Fairy-Stockfish/pull/1034
I tested it via Jocly on a modified version of Timurid (12x10). And it all
seems to work.

PR #1034, opened 2026-08-31
Bent riders: griffon, rhino, ship, snake, osprey

fhoudebert:
Adds a `y` prefix to the Betza parser for **bent riders**: a piece that takes
one leg, may stop there, and otherwise rides on without limit 45 degrees off
that leg, away from where it started. Five pieces come out of it, all
declarable through `customPieceN`, none of which is expressible today:

| Betza | piece | move |
|---|---|---|
| `yF` | griffon | diagonal step, then an orthogonal ride out |
| `yW` | rhino | orthogonal step, then a diagonal ride out |
| `yvF` | ship | the griffon, keeping only the file-wise ride |
| `vyW` | snake | the rhino, stepping only along a file first |
| `yD` | osprey | two-square orthogonal leap, then a diagonal ride out |

Two commits on top of `6d9d0f5`, 361 insertions across 10 files:
`bf66fa7 add bent riders` and `c18ed4d bent riders : add osprey`.

Related: #962, where @HGMuller implemented a superset of this on the `pieces`
branch of his repository, and #966 / #967.

`PieceInfo` currently has only `steps`, `slider` and `hopper`, none of which
can encode a change of direction mid-move, so this is a data-model addition
rather than a parser tweak. What made it worth doing now: the crashes
@HGMuller reported in #962 turned out to be the #966 bug you fixed in #967, so
there is no known blocker left on top of master.

## How the attacks are computed

**No new magic table.** A griffon's relevant-occupancy mask spans two whole
files and two whole ranks — around 24 squares on 8x8 before edge trimming, and
worse on 12x10 — so a dedicated magic would be memory-hungry for a piece most
variants never use. The attacks are composed from the rook and bishop magics
instead.

The construction rests on one observation about `QueenDirections`, which this
patch adds in **rotational order**, so that stepping ±1 through the table is a
45 degree turn and the diagonals sit at the odd indices:

> Rotating by ±1 flips the index parity. So **both** continuations of an
> orthogonal leg are diagonal, and **both** continuations of a diagonal leg
> are orthogonal — never a mix.

That is what makes a single lookup enough. For each leg:

1. Step or leap to the corner square `t`. If it is off the board, wraps
   (`distance(s, t) != len`), or is occupied, this leg contributes nothing
   further — an occupied corner blocks the ride, which is the rule.
2. Take **one** attack lookup from `t`: `attacks_bb<ROOK>` when the leg index
   is odd (diagonal leg, orthogonal continuations), `attacks_bb<BISHOP>` when
   it is even. One probe covers both continuations at once, and it already
   applies the blocking rules along them, which are exactly a bent rider's.
3. Mask with `RayBB[(i+1)&7][t] | RayBB[(i+7)&7][t]` to keep only the two
   outgoing rays, discarding the two that point back toward the origin.

```cpp
Square t = Square(s + br.len[k] * QueenDirections[i]);
if (!is_ok(t) || distance(s, t) != br.len[k] || (occupied & t))
    continue;
Bitboard rays = 0;
if (br.cont[k] & 1) rays |= RayBB[(i + 1) & 7][t];
if (br.cont[k] & 2) rays |= RayBB[(i + 7) & 7][t];
b |= ((i & 1) ? attacks_bb<ROOK>(t, occupied) : attacks_bb<BISHOP>(t, occupied)) & rays;
```

The only new table is `RayBB[8][SQUARE_NB]`, the rays to the board edge in
each queen direction: **8 KB, or 16 KB with `LARGEBOARDS`**.

Cost per attack query: four legs, one rook or bishop lookup each. Since the
rook is split into H and V magics, a griffon costs 8 magic probes against 2
for a queen; a rhino or osprey rides diagonally and costs 4. Per-quadrant
magics — mask being one partial file plus one partial rank from the corner,
hence small — are the obvious follow-up if this ever shows up in a profile.

**The rider bit carries no shape.** One `RIDER_BENT` bit covers all five
pieces; the shape lives in `BentAttacks[COLOR_NB][PIECE_TYPE_NB]`, holding per
leg a direction index, a leg length and a two-bit continuation mask. Black is
the same table with each leg turned through 180 degrees, a half-turn in the
direction table, which leaves the mask unchanged. The bit is placed **after
every magic-backed rider** and stripped by a new `MAGIC_RIDERS` mask before
the loop in `rider_attacks_bb`, which indexes `magics[]` by bit position and
would otherwise run off the end.

The second commit only changes how the corner is reached: `s + len *
QueenDirections[i]`, with `distance()` catching a leg that ran off the side.
A leg longer than one square is a **leap** — nothing between is examined —
and the parity selector still holds, because what gets rotated is the
direction the leg points in, not its length.

## Why bent riders cannot use the existing fast paths

Reachability is symmetric, but **the path is not**. A griffon on a1 reaches b5
via b2-b3-b4; a griffon on b5 reaches a1 via a4-a3-a2. The blocking sets are
disjoint. Three consequences:

- `checkSquares[pt]`, computed from the king square, assumes reversibility.
- `BetweenBB`, built as the intersection of the two reciprocal attack sets,
  gives `{c2}` for a1/b5 instead of `{a2, a3, a4}`.
- `slider_blockers()` would compute pins along the wrong squares.

Putting `RIDER_BENT` in `NON_SLIDING_RIDERS` and `ASYMMETRICAL_RIDERS` is
enough, and nothing else was needed. This matches what @HGMuller reported in
#962: declaring the new rider types non-sliding and asymmetrical was what made
his king captures go away. `NON_SLIDING_RIDERS` makes `generate_all<EVASIONS>`
fall back to `target = ~pos.pieces(Us)` and makes `legal()` skip its
`between_bb` shortcut, leaving the full `attackers_to()` at the end of
`legal()` to cover pins; `ASYMMETRICAL_RIDERS` makes `slider_blockers()`,
`attackers_to()` and `gives_check()` recompute from each candidate square
rather than reversing.

`bent_path_bb()` supplies a **directed** `between_bb(s1, s2, pt)`, traced from
the piece towards the target. `checkSquares` stays the approximate forward
set, exactly as it already is for `HORSE`, which is also in
`ASYMMETRICAL_RIDERS`; it only feeds `QUIET_CHECKS`, so the cost is a little
search quality, not correctness.

One incidental fix: `psqt.cpp` reads only `slider` and `hopper` when deciding
`isSlider`, so a piece whose movement lives in `bent` was silently classified
as a non-slider. Each continuation now counts as three quarters of a slider
direction, which puts the griffon slightly below the queen — matching its
empty-board mobility of 21.4 squares against 22.75 — and values the
half-versions at about half the full piece.

## Testing

**No regression on standard chess.** `bench` gives 6180480 with and without
the patch. Perft from the start position to depth 6 = 119060324 and Kiwipete
to depth 5 = 193690690.

**Move generation** cross-checked against an independent path-walking
reference over 600 random 12x10 positions with friendly and enemy blockers,
spread across all four shapes of the first commit: 0 disagreements. On an
empty 8x8 board from d4 the five pieces give 24, 24, 12, 14 and 16 moves, each
verified by hand as well.

**Check detection.** Two positions, same pieces, the blocker moved from b3 to
a3, in a variant with `customPiece1 = g:yF`:

| FEN | expected | result |
|---|---|---|

An implementation treating the griffon as symmetric gets both backwards.
Worth keeping as a regression test.

**The osprey leaps.** With a knight on e4 all 16 moves from d4 remain,
including d4f4 and the ride g5/h6; with the landing square f4 occupied, d4f4
stays as a capture but the ride beyond disappears.

**Stress.** 34,663 plies of random games on a `debug=yes` build in a 12x10
variant holding four bent riders at once, plus 17,721 more once the osprey was
added, plus `go depth 22` on K+G vs K endgames, which is where #962 crashed.
No assertion failures.

## Cost

Comparing like with like — both as user-defined variants, both at
`go movetime 6000` from the start position — a chess variant with a griffon in
place of the queen runs at 299 knps against 504 knps, so **about 40% slower**.
That is the asymmetrical and non-sliding paths plus the extra magic probes,
and it only affects variants that use the piece.

## Questions

1. **The first commit carries an unrelated change.** It also adds
   `enPassantTargetTypes`, a `PieceSet` letting a non-pawn leave en passant
   squares on any move of more than one square, which today only happens on a
   move its *initial* move set alone could make. I needed it for a piece that
   keeps a two-square step all game. Say the word and I will split it into its
   own PR; the two are independent.
2. **Notation.** `y` with a directional modifier before it restricting the
   first leg (`vyW`) and a single `v`/`h` straight after it restricting the
   ride (`yvF`). `yafsF` is accepted as an alias of `yF` so a piece written for
   @HGMuller's branch parses unchanged. Is that the spelling you want in the
   documented Betza subset, or should it track his branch more closely?
3. Would you rather upstream his `pieces` branch as a whole now that #967 has
   landed? It is broader — ski sliders, long-stride riders — and this PR
   should be dropped in its favour if so.
4. Is the ~40% acceptable, or would you want per-quadrant magics up front?

NNUE does not know the pieces, so variants using them run on classical eval.

### DIFF
--- a/src/bitboard.cpp
+++ b/src/bitboard.cpp
@@ -37,6 +37,9 @@ Bitboard PseudoMoves[2][COLOR_NB][PIECE_TYPE_NB][SQUARE_NB];
 Bitboard LeaperAttacks[COLOR_NB][PIECE_TYPE_NB][SQUARE_NB];
 Bitboard LeaperMoves[2][COLOR_NB][PIECE_TYPE_NB][SQUARE_NB];
 Bitboard BoardSizeBB[FILE_NB][RANK_NB];
+Bitboard RayBB[8][SQUARE_NB];
+BentRider BentAttacks[COLOR_NB][PIECE_TYPE_NB];
+BentRider BentMoves[2][COLOR_NB][PIECE_TYPE_NB];
 RiderType AttackRiderTypes[PIECE_TYPE_NB];
 RiderType MoveRiderTypes[2][PIECE_TYPE_NB];
 
@@ -186,7 +189,21 @@ namespace {
     return b;
   }
 
-  Bitboard lame_leaper_attack(std::map<Direction, int> directions, Square s, Bitboard occupied) {
+  void init_rays() {
+
+  for (int i = 0; i < 8; ++i)
+      for (Square s = SQ_A1; s <= SQ_MAX; ++s)
+      {
+          Bitboard b = 0;
+          for (Square t = s + QueenDirections[i];
+               is_ok(t) && distance(t, t - QueenDirections[i]) == 1;
+               t += QueenDirections[i])
+              b |= t;
+          RayBB[i][s] = b;
+      }
+}
+
+Bitboard lame_leaper_attack(std::map<Direction, int> directions, Square s, Bitboard occupied) {
     Bitboard b = 0;
     for (const auto& i : directions)
     {
@@ -198,6 +215,52 @@ namespace {
   }
 
 }
+
+/// bent_path_bb() returns the squares a bent rider of type 'pt' standing on
+/// 'from' has to pass through to reach 'to', excluding both end points, or 0
+/// if 'to' is out of reach or is reached by the plain step onto the corner.
+/// Unlike a straight slider, the path depends on which end the piece starts
+/// from: a Griffon on a1 reaches b5 through b2-b3-b4, while a Griffon on b5
+/// reaches a1 through a4-a3-a2. Colour is not available here, and matters
+/// only for bent riders whose legs are not centrally symmetric, so both are
+/// tried; a miss costs a pin that goes undetected, never a legality error,
+/// since bent riders are in NON_SLIDING_RIDERS and legal() recomputes.
+
+Bitboard bent_path_bb(PieceType pt, Square from, Square to) {
+
+  assert(is_ok(from) && is_ok(to));
+
+  for (Color c : { WHITE, BLACK })
+  {
+      const BentRider& br = BentAttacks[c][pt];
+      for (int k = 0; k < br.count; k++)
+      {
+          int i = br.dir[k];
+          Square corner = Square(from + QueenDirections[i]);
+          if (!is_ok(corner) || distance(from, corner) != 1)
+              continue;
+          if (corner == to)
+              return 0; // reached by the plain step, nothing in between
+          for (int j = 0; j < 2; j++)
+          {
+              if (!(br.cont[k] & (1 << j)))
+                  continue;
+              Direction d = QueenDirections[(i + (j ? 7 : 1)) & 7];
+              Bitboard b = square_bb(corner);
+              for (Square s = Square(corner + d);
+                   is_ok(s) && distance(s, s - d) == 1;
+                   s += d)
+              {
+                  if (s == to)
+                      return b;
+                  b |= s;
+              }
+          }
+      }
+  }
+  return 0;
+}
+
 
 /// safe_destination() returns the bitboard of target square for the given step
 /// from the given square. If the step is off the board, returns empty bitboard.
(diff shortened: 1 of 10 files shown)


HGMuller:
Note that this is not valid XBetza notation, and that XBoard won't understand it. I am not against making a concerted change in Betza notation in FSF and XBoard simultaneously, but I strongly recommend against doing it this way. There already exists an improvement to XBetza (in terms of understandability) partially implemented in the Interactive Diagram, known as the bracket notation. This does away with the 'a' modifier for separating legs of multi-leg moves, but uses enclosing brackets and separating hyphens or question marks, allowing each leg to specify its own Betza atom.

So a Griffon would be [F?R], meaning an F-step optionally followed by a Rook slide (by default) in the outward direction. A Griffon that could not end on the F squares would be denoted [F-R]. A rider that bents 90 degrees at the W squares would be [W?sR] if it can end on the W squares, and [W-sR] if it just has to pass through. A Tamerlane Picket would be [F-B].

Asymmetric versions can then be indicated by treating the brackets as if they are an oblique atom. E.g. the Ship would be v[F?R]. In a-notation such things are quite problematic.

fhoudebert:
I will try to propose a modified version that takes into account HG Muller's comments on the notation.

fhoudebert:
I’ve made another proposal to try to get "bent riders" added, taking your
previuous comments into account:
https://github.com/fairy-stockfish/Fairy-Stockfish/pull/1035

PR #1042, opened 2026-09-14
Feat: Support the g modifier for D and A moves

mousetail:
This adds support for `gD` and `gA` custom pieces, which jump over an exactly adjacent piece and lands on the other side. Similar to checkers though of course capturing is not supported. The "opposite" of lame leapers like the elephant which move two steps but need the intermediate square to be occupied.

The `p` modifier works identically go `g` in these cases since the piece can only move one spot past the obstacle anyways. So `pD`/`pA` is the same as `gD/gA`.


mousetail:
It seems using `pgW2` or `pgF2` also works for this case already. So maybe this change isn't strictly needed, however, it doesn't hurt. You decide

PR #1043, opened 2026-09-16
Add Spark Chess variant

GenMate:
Spark Chess is a capture-the-king variant built around two pawn-based mechanics: Shift and Swap. The rules were substantially simplified after the earlier Solomachev Chess proposal.

This PR adds the variant as `sparkchess`, with variant-specific move generation, SAN, FEN/PGN interoperability, Python and JavaScript bindings, documentation, and regression coverage. The normative rules and notation are in `SPARK_CHESS.md`.

Key rules:

- kings are capturable and there is no castling, double pawn move, or en passant;
- Shift moves a pawn to an empty coordination node;
- Swap covers a direct knight-pawn exchange and a relay exchange through a knight;
- no legal moves and the first repeated position are draws;
- the documented pawnless drawn endings are adjudicated immediately;
- chess and Chess960 castling fields are accepted on FEN import, ignored, and exported as `-`.

The implementation is guarded by the Spark Chess variant flag and does not change move generation or adjudication for existing variants. Tests cover move generation, incremental state, FEN validation, SAN, PGN import/export, JavaScript bindings, repetition, terminal results, and the documented draw rules. The standard chess bench result remains unchanged from upstream.

Interactive rules and playable examples: https://sparkchess.solomachev.club/


ianfab:
Thanks, but this is way too much variant specialized code to be maintainable in mainline Fairy-SF, so I have to reject.

Issue #804, opened 2024-06-27
Crash in variant Coffeehouse


autocorr:
There appears to be a segfault in the variant "coffeehouse" when null moves are generated. It happens frequently on the fishnet worker I use, but is also reproducible from (most) of the same positions on fairyground. The variant definition used can be found below:

```
[coffeehouse:crazyhouse]
mustCapture = true
```

An example can be reproduced on fairyground by setting `multipv` to 2, `hash` to 256 MB, `threads` to 4, and an initial move for White of `f2f3` then starting analysis. After reaching a search depth of around 9, a null/invalid move is reached in the second line and the engine subsequently crashes. Interestingly this particular move doesn't force a capture.

I don't have much more insight into the cause, except that raising the hash size tends make the errors less frequent. Reading through the current issues, it's possible that this is also related to #800 , since this definition inherits from ZH. Although I can't do it today, I can test a definition without inheritance soon to see if that works around the problem.

Edit: I tried inheriting directly from `chess` by adding the rules for Crazyhouse to the definition (see below), and it seems to have worked around the issue.

```
[coffeehouse:chess]
startFen = rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR[] w KQkq - 0 1
pieceDrops = true
capturesToHand = true
mustCapture = true
```

ianfab:
Thanks. Can you reproduce this with native Fairy-SF? So far I failed to reproduce it regardless of multipv, hash, and thread setting, even with debug=yes.

PraseodymiumSpike:
It could be related to https://github.com/fairy-stockfish/fairy-stockfish.wasm/issues/14.

PR #1045, opened 2026-09-19
Merge Stockfish 19

ianfab:
Merges official Stockfish up to SF 19 (must be merged as a merge commit to keep the upstream history).
Classical evaluation and the variant NNUE (HalfKAv2Variants, 0x7AF32F20) are kept; all existing networks
keep working.

## Structure of the branch

1. `da301c90` Merge Stockfish 19, followed by variant specific search fixes
   - `49499ad7` restrict new upstream pruning for blast variants (atomic)
   - `e1dd6726` limit singular extensions to one ply for drop variants (crazyhouse, shogi)
   - `bd05ecae` restore hybrid evaluation for check counting variants (3check)
   - `00a96135` move ordering for variants with mandatory captures (antichess, losers)
   - `a0ab72d5` fix pyffish and ffish.js source lists
   - `8882bacb` do not search root positions without legal moves (merge bug found by assert-enabled self-play)
   - `116b9e7a` fix parsing of integer options of variants.ini (Value is an alias of int since the merge,
     found by the variants.ini check of the CI)
   - `b7ec3c5c` wait for the search before changing the evaluation settings (data race of useNNUE); these two are the last commits of the branch
   - `f9040992` index pawn history by compact piece slot (memory per thread back to the level of master)
2. NNUE code structure of upstream, no functional change (node-identical for 14 variant/net combinations,
   192 of 196 tested networks load with identical raw output, the other 4 do not load on master either)
   - `a6b53acc` accumulators move from StateInfo to an AccumulatorStack in the worker
   - `9cc080cf` Network class owned by the engine (NUMA replicated), `evaluate_nnue.*` removed
   - `d585c786`, `cb6e0840` layer implementations and SIMD header of upstream
   - `e0460080` no static limit of input dimensions (large networks load in non-largeboard builds)
   - `5d78e936` accumulator caches (Finny tables)
3. New network support
   - `9a7792d4` networks without king buckets (A layout) for all variants, detected by the file size
   - `3c2e4f87` NNUE classes parametrized by the network architecture, selected at runtime by the file header
   - `f0250ade` networks of USI shogi engines in the format of YaneuraOu (HalfKP, 256x2-32-32 and
     768x2-16-64, e.g. Suisho5 and AobaNNUE), largeboard builds only, selected by file name `shogi-*.bin`.
     Output bit-identical to YaneuraOu on 900+ positions per network.

## User visible changes

- `Slow Mover` is gone (removed upstream).
- Info strings when a network is loaded, as upstream.
- No new options.

## Strength vs master (variantfishtest, 5s+0.05, 1 thread)

NNUE, same network on both sides: crazyhouse +151, shogi +93, kingofthehill +72, horde +65, chess +34,
xiangqi +30, atomic +26, racingkings +25, makruk +17, janggi +17, 3check +7, antichess -14 (±33).

Classical: horde +133, racingkings +46, chess +44, kingofthehill +44, crazyhouse +33, makruk +13, 3check +10,
xiangqi -13, janggi -17, atomic -20, shogi -37, antichess -45 and losers -68 (both before `00a96135`, which
gained about +20 in each).

LTC 20s+0.2 and 4 threads were checked for plausibility only, without anomalies.

## Known issues

- Remaining regressions with classical evaluation in shogi, antichess and losers, see above.
- USI shogi networks: the evaluation is used as is (divided by the scale of the architecture), so the
  displayed centipawns follow the normalization of Fairy-Stockfish and not the one of USI engines, and
  the scale is not tuned for the search of Fairy-Stockfish.
- Not verifiable locally, to be checked by CI: Windows (MSVC/MinGW), macOS, ARM, 32-bit, ffish.js, wheels.
- Already present on master and unchanged: debug assertions in crowdedxiangqi, xhess/leaperhouse,
  xiangqihouse and 1d-chess, the XBoard hang on `quit` while searching, and the stack usage of #804.

Bench: 2982107

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Torom:
The `compiler` output looks a little odd compared to upstream Stockfish:
```
Fairy-Stockfish -20260920-b2ad0e35 by Fabian Fichter

Compiled by g++ (GNUC) 16.2.1 on Linux
Compilation architecture   : x86-64-modern
Compilation settings       : 64bit SSE41 SSSE3 SSE2 POPCNT
__VERSION__ macro expands to: 16.2.1 20260810
```
upstream stockfish
```
Stockfish dev-20260919-17a6c8f1 by the Stockfish developers (see AUTHORS file)

Compiled by                : g++ (GNUC) 16.2.1 on Linux
Compilation architecture   : x86-64-avx512icl
Compilation settings       : 64bit AVX512ICL VNNI AVX512 BMI2 AVX2 SSE41 SSSE3 SSE2 POPCNT
Compiler __VERSION__ macro : 16.2.1 20260810
```
And perhaps more importantly, native architecture detection does not work yet. `x86-64-modern` is used instead of `x86-64-avx512icl`.

ianfab:
[Claude] Thanks, all three were places where the merge kept the old Fairy-Stockfish side without a reason. Fixed in 890f3662: `ARCH` now defaults to `native` as in Stockfish 19 (the detection itself worked, but only with an explicit `ARCH=native`), the compiler info uses the upstream labels, and dev builds are named `dev-<date>-<sha>`, which also removes the stray hyphen in the version string.

RainRat:
_I asked Sol to review this PR and it posted a comment before I could review it. It's already up, so I'm just leaving the rest as is._

I did a second, deeper correctness pass over the current PR head, using separate read-only reviews for variant/state logic, search/move generation, and engine/protocol integration. I limited this to reproducible failures in built-in variants or variants shipped in `variants.ini` (no hardening, performance work, or speculative option combinations).

Two findings already have focused PRs and should be included here:

- [#1020](https://github.com/fairy-stockfish/Fairy-Stockfish/pull/1020): multiple en-passant targets on one file currently XOR the same file key twice and cancel, so Berolina positions with `e3e4` rights hash identically to the same position with no EP rights.
- [#1040](https://github.com/fairy-stockfish/Fairy-Stockfish/pull/1040): Atomic Duck generates wall destinations from pre-explosion occupancy and has matching legality errors. Explosion-vacated squares are valid wall destinations.

I also found the following definite issues not covered by those PRs.

### 1. Atlantis rejects legal wall-only check evasions

`[atlantis:chess]` uses `wallingRule = edge` and `wallOrMove = true`. `Position::legal()` checks the king against occupancy that omits the newly placed wall.

```text
./stockfish load variants.ini
setoption name UCI_Variant value atlantis
position fen 7k/8/8/8/8/8/8/K6r w - - 0 1
go perft 1
```

The current head returns only the two king moves. It should also allow the six wall-only evasions `a1a1,a1b1` through `a1a1,a1g1` (8 nodes total).

### 2. Alapo can award a flag win when every qualifying piece is attacked

`flag_reached()` checks `flagPieceSafe` only while the number of pieces in the flag region is *exactly* `flagPieceCount`. With extra pieces, it skips safety entirely. In shipped Alapo, this position incorrectly awards White the flag win even though both white goal rooks are attacked:

```text
RR4/6/6/rr4/6/6 b - - 0 1
```

The check needs to count safe pieces and require at least `flagPieceCount`; unsafe extras can still be ignored.

### 3. Pseudo-royal check is missing from draw adjudication and quiescence search

For variants such as built-in Petrified, `checkers()` is empty when the commoner is in pseudo-royal check.

- The n-move/counting-rule guards therefore allow a draw to override pseudo-royal checkmate. `1r6/8/8/8/8/2k5/r7/K7 w - - 100 1` has no legal move and is mate, but `is_optional_game_end()` currently claims the n-move draw.
- Qsearch treats the same kind of checked position as a stand-pat position and searches captures only. From `1r6/8/8/8/8/r1k5/8/K7 b - - 0 1`, a depth-1 search currently misses Black's quiet rook mates (for example `a3a2`).

The minimal fix is to include `checked_pseudo_royals(side_to_move())` in the search/adjudication notion of being in check, and let qsearch generate all legal pseudo-royal evasions.

### 4. Several rule-relevant states are missing or incorrectly updated in Zobrist keys

These cause real TT/repetition aliases in supported variants:

- **Ataxx free drops:** DROP hashing indexes `inHand[count - 1]`, but `freeDrops` deliberately keeps the hand count at zero. Every Ataxx clone/drop reads `inHand[-1]` (undefined behavior) and produces an invalid incremental key.
- **Seirawan gating:** adding the gated piece does not update the pocket portion of the main key, and the material key uses the post-insertion count rather than `count - 1`. Incrementally playing `b1c3e` does not produce the same state keys as parsing its resulting FEN.
- **Gate rights:** `gatesBB` affects legal moves but is absent from the main key. A Seirawan position with gate rights hashes like the same board/side/pockets without them.
- **Promotion provenance:** promoted/unpromoted provenance is absent from the key. In Crazyhouse, `Q` and `Q~` on the same square hash alike even though capturing them yields different pocket pieces; Kyoto-style demotion has the analogous problem.
- **Placement castling:** when a dropped king or rook creates castling rights, `do_move()` changes `castlingRights` without applying the corresponding castling-key transition.
- **Janggi passes:** `st->pass` determines whether the next pass ends the game, but is absent from the key. A position reached after one pass hashes like the same FEN parsed directly, although a pass has different results in the two states.

I fixed this locally by guarding free-drop hand hashing, correcting Seirawan's pocket/material transitions, updating dynamically created castling rights, and adding variant-state keys for gates, promotion provenance, and the prior-pass flag. The extra state calculation is restricted to variants that use those rules.

### 5. Root Syzygy ranking is applied to non-chess variants

`ThreadPool::start_thinking()` calls `rank_root_moves()` for every variant, while the deeper probe correctly checks `UCI_Variant == "chess"`. With KQvK tables loaded in King of the Hill, for example:

```text
7k/8/8/8/8/1qK5/8/8 w - - 0 1
```

`c3d4` is an immediate KOTH win, but standard-chess tablebase ranking places `c3b3` (a standard draw after capturing the queen) in a higher root rank and can exclude the variant-winning move from search. The chess-only guard belongs inside `Tablebases::rank_root_moves()` so every caller is protected.

### 6. Protocol positions retain a dangling worker pointer after thread resize

`Engine::set_position()` and XBoard `setboard()` attach the protocol `Position` to `main_thread()->worker`. Changing `Threads` destroys and recreates the worker pool without rebinding that position. A later capture dereferences the stale pointer when prefetching the material table.

Concrete UCI sequence:

```text
position fen 4k3/8/8/8/8/8/4p3/4R1K1 w - - 0 1
setoption name Threads value 2
go perft 2
```

Protocol positions can safely use a null worker (the search copies receive their own workers, and the capture path already permits null for perft/bindings).

I implemented focused fixes and regressions for all of the above locally. Final verification:

- forced `largeboards=yes all=yes` release build
- assertion-enabled build
- ASan+UBSan reproductions for Ataxx free drops and thread resizing
- all 25 Python binding tests
- `tests/perft.sh all`
- `tests/protocol.sh`
- `tests/reprosearch.sh`
- `stockfish check variants.ini`
- standard bench completed

One initially reported JS/root terminal-state concern was disproved because `generate<LEGAL>` already returns no moves after an immediate variant result; I excluded it from this list.

ianfab:
As far as I understand from a superficial read Sol reviewed the branch, not the diff/commits. Fixing unrelated bugs that are not introduced by the PR is bad practice. If there are bugs specifically related to the PR please report them here. For other bugs please use separate issues or the respective PRs to discuss them.

ianfab:
Bughouse regression test of this PR against master, using [fairyfishtest](https://github.com/ianfab/fairyfishtest):

- PR head `0069d339` vs. master `9f778da6`, both built with `ARCH=x86-64-bmi2`
- bughouse, 10+0.1, concurrency 1, start positions from `bughouse-p4-f12.epd`
- SPRT [-15, 0], stopped manually before it concluded

```
Total: 202 W: 82 L: 91 D: 29 ELO: -15.49 +-44.6 (95%) LOS: 24.7% LLR: -0.23 (-2.94,2.94) [-15.00,0.00]
```

The result is inconclusive. There is no clear regression in bughouse, but 202 games cannot rule out a loss of up to about 30 Elo either. The machine was under some unrelated background load, which adds noise but should not favour either side.

Torom:
`fairyfishtest.py ../Engines/fairy-stockfish_pr_avx512 ../Engines/fairy-stockfish_master_avx512 -v bughouse -n 30000 -t 30+0.3 -c 8 -b bughouse-p4-f12.epd -s --elo0 -15 --elo1 0` produced an illegal move:
```python
match3: Illegal move: e2e3
match3:b1e2:reader: Engine rejected a command: Illegal move: a3b4
match3:board2: Board 2 failed
ValueError: Invalid move 'e2e3'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/torsten/chess/fairyfishtest/fairyfishtest.py", line 665, in play_board2
    game2.play()
    ~~~~~~~~~~^^
  File "/home/torsten/chess/fairyfishtest/fairyfishtest.py", line 437, in play
    self.partner.set_holdings(captured)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/torsten/chess/fairyfishtest/fairyfishtest.py", line 499, in set_holdings
    holdings = self.get_holdings()
  File "/home/torsten/chess/fairyfishtest/fairyfishtest.py", line 493, in get_holdings
    fen = sf.get_fen(self.variant, self.get_start_fen(), self.moves)
SystemError: <built-in function get_fen> returned a result with an exception set
match3: Match failed
ValueError: Invalid move 'e2e3'
```

I also notice quite a few time losses despite the time control with increment.

ianfab:
Thanks. Can you please provide more detail? Like is this a regression compared to master, and do you have debug output of which situation this occurs in?

Until the fixes in both fairyfishtest and Fairy-SF yesterday bughouse testing always has been very unstable and just barely good enough to get an indication of whether patches make sense. Therefore it is no surprise that it still does not work perfectly, so in the scope of this PR it is only relevant if what your test shows is a regression compared to master.

Torom:
A run of master vs. master also stopped after 43 games with an illegal move. So I don't think it's a regression.

Torom:
A match PR vs. PR runs fine for 300 games. So there's a good chance that this actually improves stability compared to master.

Torom:
The `native` architecture does not recognize `dotprod` correctly and builds just `armv8` instead of `armv8-dotprod`. Even though `get_native_properties.sh` outputs `armv8-dotprod`.

ianfab:
[Claude] Thanks for testing. I cannot reproduce this on x86, and I do not see a difference to Stockfish 19 in that path: the `native` resolution in the Makefile (`$(SHELL) ../scripts/get_native_properties.sh | cut -d " " -f 1`), the script itself and the `armv8-dotprod` block with `-march=armv8.2-a+dotprod -DUSE_NEON_DOTPROD` are identical to upstream. The only ARM-related deviation is the missing universal binary support, which is planned for a separate PR.

To narrow it down, could you share:

1. platform and compiler (Linux/Android/macOS, gcc or clang version, `COMP=` if set),
2. the output of `make ARCH=native config-sanity`, which shows whether `ARCH` is already resolved to `armv8` or whether only `dotprod` gets lost,
3. the "Compilation architecture" and "Compilation settings" lines printed by the built engine (`./stockfish compiler`).

One thing to keep in mind is that make runs the script with `/bin/sh` rather than your interactive shell, so if the detection differs between the two it would show up in the config-sanity output.

Torom:
Sorry, it's totally my fault. My old `makefairystockfish` script defaulted to `armv8` because there wasn't an `armv8-dotprod` available. If I just run `make profile-build`, I get an `armv8-dotprod` binary.

